PR #389, opened 2021-04-11
Demo-001

senkiv-oleh:
- [DEMO LINK](https://senkiv-oleh.github.io/layout_miami/)


src/styles/blocks/contact.scss
```diff
@@ -0,0 +1,80 @@
...
+    font-size: 14px;
+    line-height: 27px;
+    color: #9fadc2;
+    font-family: Roboto, sans-serif;
+  }
+
+  &__phone {
+    font-size: 16px;
+    line-height: 27px;
+    color: #3e3e3e;
```
varseniuk, line 58:
```suggestion
    color: #3e3e3e;
```
Better to create variables for font colors and font sizes. For example, you use this color three times in your cod. It would be easy to use meaningful variable than this color code

PR #431, opened 2021-04-13
v1

vitalya2626:
[DEMO LINK](https://vitalya2626.github.io/layout_miami/)


src/styles/blocks/info.scss
```diff
@@ -0,0 +1,63 @@
+.main__info {
```
varseniuk, line 1:
```suggestion
.main__info {
```
'main__info' is an element of block `main` (as well as `main__contacts`). These styles should be n another file. Nove them to `main.scss`. Or it would be better to create a new file `main.scss` for the `main` block and rename your current `main.scss` to `styles.scss`

varseniuk (changes requested):

varseniuk (approved):
🚀

PR #490, opened 2021-07-18
Develop

ihor-jpeg:
-   [DEMO LINK](https://ihor-jpeg.github.io/layout_miami/)


sasha-morozov (approved):
Imho, everything looks ok to me. It seems that everything has been done according to the video instructions, so I have nothing to add.

src/styles/blocks/container.scss
```diff
@@ -0,0 +1,9 @@
+.container {
+  margin: 0 auto;
```
varseniuk, line 2:
```suggestion
```
Don't use outer spacings for BEM blocks. Use mix of classes. `.container` is a BEM block

src/index.html
```diff
@@ -1,13 +1,418 @@
...
+                  <br>
+                  suite 700
+                  <br>
+                  Minneapolis, MN 55401
+                </a>
+              </div>
+            </div>
+
+            <div class="contacts__item grid__item--8-12">
+              <form action="#">
```
varseniuk, line 380: Make all fields in the form required. Now we can send an empty form

varseniuk (approved):
All is great. 
Pay attention to margins in the BEM-blocks and make all fields in form required

PR #500, opened 2021-07-20
first commit

Moroz-Dmytro:
[DEMO LINK](https://Moroz-Dmytro.github.io/layout_miami/)


alex-glotov67 (approved):
Looks great! Hover effects for the gallery are awesome!

src/styles/utils/_mixins.scss
```diff
@@ -1,6 +1,11 @@
-@mixin hover($_property, $_toValue) {
-  transition: #{$_property} 0.3s;
-  &:hover {
-    #{$_property}: $_toValue;
+@mixin phoneLandscape {
+  @media (max-width: 900px) and (orientation: landscape) {
```
TarasHoliuk, line 2: nice solution
didn't know that we can use `orientation: landscape`

src/index.html
```diff
@@ -1,13 +1,451 @@
...
+                  suite 700
+                  <br>
+                  Minneapolis, MN 55401
+                </a>
+              </div>
+            </div>
+
+            <div class="grid__item grid__item--8-12">
+              <form action="">
+                <input
```
varseniuk, line 413: Nake all fields in this form required. Now we can send an empty form

src/styles/blocks/container.scss
```diff
@@ -0,0 +1,9 @@
+.container {
+  margin: 0 auto;
```
varseniuk, line 2:
```suggestion
```
Don't use outer spacings (margins) for BEM blocks. `.container` is a block here

src/styles/blocks/place.scss
```diff
@@ -0,0 +1,49 @@
+.place {
+  &__photo-container {
+    position: relative;
+    width: 100%;
+    max-height: 256px;
```
varseniuk, line 5: According to the design, on small screens, these pictures should become smaller (height 131px). You have a height of 256px for all widths.

varseniuk (changes requested):
Great job! Let's make some small fixes and it's done

src/styles/blocks/place.scss
Moroz-Dmytro: They become smaller on small screens, this max-height doesn`t let them become higher than 256px, because on small screens it looks bad

src/styles/blocks/container.scss
Moroz-Dmytro: In video, Misha said that we can use this container with margins in this case, but we should remember that we can`t use this block like BEM-block.  We use it to set max-width for all blocks and to set them in the middle

varseniuk: Ok

src/styles/blocks/place.scss
varseniuk: Really. My bad. Haven't noticed. Sorry

varseniuk (approved):
Great job! 🔥

PR #502, opened 2021-07-20
Miami

nastyyvolk:
[DEMO LINK](https://nastyyvolk.github.io/layout_miami/)


alex-glotov67 (changes requested):
All great! But if you choose iphone SE or some another mobile device, the menu dropdown is in a horizontal position. Maybe you should make media query for min-hight for menu, and make it in two column or some your ideas)

danheim (approved):

PR #620, opened 2021-09-06
miami

lilia-kuleba:
- [DEMO LINK](https://lilia-kuleba.github.io/layout_miami/)


aliceblunt8 (changes requested):
make scale not that big and add it only to photos

image

aliceblunt8 (approved):

PR #686, opened 2021-10-26
Miami

mariia-bila:
https://mariia-bila.github.io/layout_miami/


aionova (approved):
- all the links that open maps, should open them in a new tab
- that scale is too big. I would make it 1.1.
Screenshot 2021-11-04 at 13 23 44
- add transition and cursor: pointer to all the buttons

PR #688, opened 2021-10-26
Miami v1.0

andriibulbuk:
- [DEMO LINK](https://kurtkobein.github.io/layout_miami/)


aionova (approved):
- address should open in a new tab
Screenshot 2021-11-04 at 08 49 25
- it would be nice if indent before menu items would be the same as indent before phone number.
Screenshot 2021-11-04 at 08 50 15
- on the portrait mode of iphone 5 I can't see all the items of the menu
Screenshot 2021-11-04 at 08 51 04

PR #696, opened 2021-10-27
Miami landing

MagicBox15:
- [DEMO LINK](https://MagicBox15.github.io/layout_miami/)

### DIFF
--- a/src/styles/blocks/button.scss
+++ b/src/styles/blocks/button.scss
@@ -15,7 +15,10 @@
   color: $colorBlue;
 
   cursor: pointer;
-  transition: all $duration;
+  transition: color $duration,
+    border-color $duration,
+    background-color $duration,
+    box-shadow $duration;
 
   &--primary {
     color: #fff;
--- a/src/index.html
+++ b/src/index.html
@@ -438,18 +438,20 @@
             </div>
 
             <div class="grid__item grid__item--8-12">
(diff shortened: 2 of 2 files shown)


src/styles/blocks/button.scss
```diff
@@ -0,0 +1,40 @@
...
+  border-radius: 8px;
+
+  font-weight: bold;
+  font-size: 16px;
+  line-height: 14px;
+  text-decoration: none;
+  color: $colorBlue;
+
+  cursor: pointer;
+  transition: all $duration;
```
aionova, line 18: indicate which property you use transition for. Don't use transition all

src/styles/blocks/form-field.scss
```diff
@@ -0,0 +1,31 @@
+.form-field {
+  display: block;
+  padding: 14px 16px;
+  height: 48px;
+  width: 100%;
+  border: 2px solid transparent;
+  border-radius: 8px;
+  color: #000;
+  background-color: #f7f8ff;
```
aionova, line 9: use variables for colors

aionova (changes requested):
- when I submit the form, the shouldn't be an error.
- form shouldn't submit, if not all the fields are filled

aionova (approved):
GJ!

PR #699, opened 2021-10-28
Miami lending v1

vlad-tuzenko:
[Demo](https://vlad-tuzenko.github.io/layout_miami/)

### DIFF
--- a/src/styles/blocks/about.scss
+++ b/src/styles/blocks/about.scss
@@ -7,7 +7,7 @@
     align-items: center;
     letter-spacing: -1.5px;
 
-    color: #000;
+    color: $c-black;
 
     @include onDesktop {
       margin-bottom: 0;
--- a/src/styles/blocks/button.scss
+++ b/src/styles/blocks/button.scss
@@ -15,25 +15,28 @@
   color: $c-blue;
   cursor: pointer;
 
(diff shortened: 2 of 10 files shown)


src/styles/blocks/about.scss
```diff
@@ -0,0 +1,46 @@
+.about {
+  &__title {
+    margin: 0 0 48px;
+    font-weight: bold;
+    font-size: 36px;
+    line-height: 36px;
+    align-items: center;
+    letter-spacing: -1.5px;
+
+    color: #000;
```
aionova, line 10: colors should be stored in variables

src/styles/blocks/button.scss
```diff
@@ -0,0 +1,39 @@
...
+  border: 2px solid $c-blue;
+  border-radius: 8px;
+
+  font-weight: bold;
+  font-size: 16px;
+  line-height: 14px;
+  color: $c-blue;
+  cursor: pointer;
+
+  transition: all 0.3s;
```
aionova, line 18: don't use transition all. indicate which properties you use transition for

aionova (changes requested):
- form shouldn't be submitted, if some of the fields are not filled.

aionova (approved):
LGTM!

PR #702, opened 2021-10-28
Miami was created

4qwerty:
- [DEMO LINK](https://4qwerty.github.io/layout_miami/)

### DIFF
--- a/src/styles/blocks/page.scss
+++ b/src/styles/blocks/page.scss
@@ -36,7 +36,7 @@
   }
 
   &__section-title {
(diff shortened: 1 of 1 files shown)


ihor-jpeg (changes requested):
it seems like some styles are missing. the heading is too close to the section itself
Screenshot 2021-11-01 at 11 45 55
Screenshot 2021-11-01 at 11 46 26

aionova (approved):
Good job!

PR #708, opened 2021-10-28
Develop

PavloKhanas:


aionova (changes requested):
- When I press on the logo in header or footer, there shouldn't be an error.
- Form shouldn't submit when some of the fields are not filled

PR #711, opened 2021-10-28
Landing first

ImIkator:
[Demo link](https://imikator.github.io/layout_miami/)

### DIFF
--- a/src/styles/_variables.scss
+++ b/src/styles/_variables.scss
@@ -1,4 +1,4 @@
-$headerTopPadding: 48 px;
+$headerTopPadding: 24px;
 
 $color-blue: #0075ff;
 $color-blue-active: #005ecc;


src/styles/blocks/about.scss
```diff
@@ -0,0 +1,45 @@
+.about {
+  &__title {
+    margin: 0 0 48px;
+    font-weight: bold;
+    font-size: 36px;
+    line-height: 36px;
+    align-items: center;
+    letter-spacing: -1.5px;
+    color: #000;
```
aionova, line 9: all the colors should be put in variables

src/styles/blocks/testimonials.scss
```diff
@@ -0,0 +1,66 @@
...
+      margin-bottom: 0;
+    }
+
+    @include onDesktop {
+      margin-bottom: 0;
+    }
+  }
+}
+
+.testimonial {
```
aionova, line 15: there should be a separate file for each BEM-block

aionova (approved):
Nice!
- these should be some place between the logo and the top of the page
Screenshot 2021-11-01 at 11 11 36

src/styles/blocks/testimonials.scss
ImIkator: You're right, thank's, done.

src/styles/blocks/about.scss
ImIkator: Correct, corrected)

PR #717, opened 2021-10-29
implemented miami page

tmrsl:
- [DEMO LINK](https://tmrsl.github.io/layout_miami/)

### DIFF
--- a/src/styles/blocks/button.scss
+++ b/src/styles/blocks/button.scss
@@ -15,7 +15,7 @@
   color: $c_blue;
 
   cursor: pointer;
-  transition: all $transition-duration;
+  transition: color, border-color, background-color, box-shadow $transition-duration;
 
   &--primary {
     color:$c_white;
--- a/src/index.html
+++ b/src/index.html
@@ -27,7 +27,7 @@
       <div class="container">
         <div class="header__content">
           <div class="top-actions">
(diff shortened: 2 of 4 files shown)


ihor-jpeg (changes requested):
there a few things to work on.
first, i can still get to the images while in the menu using the `TAB` key
Screenshot 2021-11-01 at 11 57 27
second, i can submit an empty form. it's a no-no
third, i think the images get too large on hover. here, the image goes beyond the viewport and covers the text
Screenshot 2021-11-01 at 11 58 02
Screenshot 2021-11-01 at 11 58 11

src/styles/blocks/button.scss
```diff
@@ -0,0 +1,38 @@
...
+  border-radius: 8px;
+
+  font-weight: bold;
+  font-size: 16px;
+  line-height: 14px;
+  text-decoration: none;
+  color: $c_blue;
+
+  cursor: pointer;
+  transition: all $transition-duration;
```
aionova, line 18: indicate which properties you use transition for

aionova (approved):
- form shouldn't be submitted if some of the fields are not filled

aionova (changes requested):
.

ihor-jpeg (approved):
Well done!

PR #833, opened 2022-01-10
v1

ov3rst33r:
...


aliceblunt8 (changes requested):
add demo link and deploy your project

aliceblunt8 (changes requested):
- fix this when scrolling menu
image
- these links should open in a new tab
image
- remake this section according the design (photos are too big)
image

also check all points from this checklist

PR #867, opened 2022-02-01
tablet syles

Marta1993-droid:


Ihor-Karpyn (changes requested):
https://www.loom.com/share/ac501112450040898ae93c76db637def

Ihor-Karpyn (changes requested):
- add cursor pointer to btn
- decrease input's paddint, compare with markup
- page broken on 320px
- strech input to 100% on mobile devices

PR #965, opened 2022-03-30
fixed

mikhail-shchuka:


marina-tilniak (changes requested):
- checklist point 8
- form should be cleared, when button pressed 
- add hover effects for this blocks 
image

aliceblunt8 (approved):

PR #976, opened 2022-04-05
Initial commit - WiP

rHashTag:
Bose Landing - WiP


src/index.html
```diff
@@ -3,11 +3,311 @@
...
+    <link
+      rel="shortcut icon"
+      href="images/favicon.ico"
+      type="image/x-icon"
+    >
     <script src="scripts/main.js" defer></script>
   </head>
   <body>
-    <h1>Miami</h1>
+    <header class="header">
```
meforoma, line 16: header should cover the whole viewport at least in tablet and laptop resolutions
cover full screen on laptop and tablet

src/index.html
```diff
@@ -3,11 +3,311 @@
...
+                  $ 129.00
+                </p>
+              </div>
+            </div>
+
+          </div>
+        </div>
+      </section>
+
+      <section class="browse-bose" id="gallery">
```
meforoma, line 127: >8	Pictures in Gallery should increase on hover

src/index.html
```diff
@@ -3,11 +3,311 @@
...
+      href="images/favicon.ico"
+      type="image/x-icon"
+    >
     <script src="scripts/main.js" defer></script>
   </head>
   <body>
-    <h1>Miami</h1>
+    <header class="header">
+      <div class="header__wrapper">
+        <div class="header__menu">
```
meforoma, line 18: once covered on lesson, please implement
>17	Fix menu for small screens (if there is not enough space for all the menu items)
>18	Disable page scrolling under the menu

src/index.html
```diff
@@ -3,11 +3,311 @@
...
+
+      <section class="contact-us" id="contacts">
+        <div class="contact-us__wrapper">
+          <h2 class="contact-us__title title">
+            Contact us
+          </h2>
+
+          <div class="contact-us__content-wrapper">
+            <div class="contact-us__submit">
+              <form
```
meforoma, line 242: >15	Page shouldn't be reloaded on form submit
(`action="#" method="get"` OR https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)

meforoma (changes requested):

src/index.html
rHashTag: Done!

src/index.html
rHashTag: 17: Menu fits in all resolutions, even down to 320 x 640. Point 17 thus not implemented.

18: Done! What I have noticed though, is that the pointer shifts to the right by the scrollbar with (18px I believe). I'm wondering is there a way to compensate for this other then changing right: 18px?

src/index.html
rHashTag: Done.

src/index.html
rHashTag: Done!

src/index.html
meforoma: @rHashTag your form still reloads the page
use .js to make sure it's not

```
const form = document.querySelector('<form selector>');

  event.preventDefault();
  form.reset();
});
```

meforoma (changes requested):
finalize the form behavior
and for consistency sake - full screen header for mobile resolutions would also be good
mobile 100vh

src/index.html
rHashTag: Done!

Also updated the 100vh for mobile and tablet views, just to be consistent.

meforoma (approved):
way to go 👍

PR #1019, opened 2022-04-16
Develop

Nazar-Cheredaryk:
where can i get demo link for this task ?)


aliceblunt8 (changes requested):
look inside readme, you need to update it with your account name

Nazar-Cheredaryk:
> look inside readme, you need to update it with your account name

but there's no demolinks 
image

PR #1036, opened 2022-04-25
My bike

DimaK123:
https://dimak123.github.io/layout_miami/


marina-tilniak (changes requested):
Good job 👍  Fix checklist point 6 and 18

PR #1053, opened 2022-05-20
New develop

yanaost:
https://yanaost.github.io/layout_miami/

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -3,12 +3,20 @@
   <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>Miami</title>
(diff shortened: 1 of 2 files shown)


jstmpelowycz (changes requested):
Good job!
Make sure you fix the following issues:

- go through the checklist one more time
- on the menu, the phone number link text looks incorrect

Awesome hover effects 🔥

mykola-koval (changes requested):
- these links should not open the page in a new tab
image
image
image

 - i would not use these images in black & white and just increased their size on hover
image

 - submitting the form should not reload the page, ideally the form should be cleared on submit

yanaost:
I prefer to leave testimonials images in black & white with filter on hover

Olexandr6 (approved):

PR #1061, opened 2022-05-22
Develop

khrystyna-sabat:
- [DEMO LINK](https://khrystyna-sabat.github.io/layout_miami/)


src/styles/blocks/recommended.scss
```diff
@@ -0,0 +1,54 @@
...
+
+  &__photo-container {
+    display: grid;
+    grid-template-columns: 4, 1fr;
+    column-gap: 24px;
+  }
+
+  &__photo {
+    width: 100%;
+    grid-column: 2/4;
```
stanislav-topikha, line 26: Why not to reuse existing grid system?

stanislav-topikha (changes requested):
Page looks good, responsive 🔥, but there is a problem with checklist requirements

Favicon isn't added 
image

You should disable scroll for page when menu is opened and add menu scroll for small screens
image
image

Add target blank for this 
image

Let's think about privacy, try to use other method.
All fields should be required by default.
Page shouldn't be reloaded on form submit
image

Recheck requirements.

stanislav-topikha:
15. Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)

stanislav-topikha (approved):

jstmpelowycz (changes requested):
Good job, the largest part of the landing is done, consider the following issues:
    - overall
    	- only the desktop version should have hover effects, on mobile/tablet those are useless and redundant
    - [**header**]
        - low-quality icons, consider downloading vector versions of images
    - [**menu**]
        - links underline hover effect is different from what is in the mockup (should be based on link content size)
        - number and `call to order` adhere to the container, but links above, like `Contacts`, have padding
Screen Shot 2022-05-24 at 15 48 17
    - [**recommended**]
        - add pointer cursor to images
    - [**contact-us**]
        - textarea should be a required field as well
        - form doesn’t get cleared on submit
        - add hover effect for the links (email, phone, address)

jstmpelowycz (approved):
Looks awesome 🔥 🚀

PR #1063, opened 2022-05-23
layout_miami created

AntonYegorov:


jstmpelowycz (changes requested):
GJ, but make sure you fix the issues:
- make sure every element have `cursor: pointer` on hover (looks more interactive)
- [**form**]:

    - there is no visual response when the form gets submitted
    - fix form fields not being required

mykola-koval (approved):
image
make this link open in a new tab

and let the form clear when form submits

PR #1093, opened 2022-05-25
Miami landing (My bike): first draft

MariiaNH:
- [MY Bike](https://mariianh.github.io/layout_miami/)


jstmpelowycz (changes requested):
Good job!
Make sure you fix the following issues:

- the checklist: 1, 2, 4, 13-15, 18
- add pointer cursors where needed (as if to make hover effects look more interactive)
- borders of some form fields and  look different from what is in the mockup

PR #1108, opened 2022-05-26
made miami landing

Victance:
-[DEMO](https://victance.github.io/layout_miami/)


jstmpelowycz (approved):
Looks awesome!
There is a small issue – images scale on hover is extensive, reduce your scale value
Also, I have a suggestion – you could try making your form get cleared on submit, that'll enhance UX

PR #1147, opened 2022-06-13
Miami

bogdana18:
[DEMO LINK](https://bogdana18.github.io/layout_miami/)


aliceblunt8 (approved):
good job!

PR #1159, opened 2022-06-16
myBikeLandingPagev1.0

fenderukr:
https://fenderukr.github.io/layout_miami/

It's not clear what background-color or image I should use for other blocks except for the menu and first one

Also, I am not sure that I properly set the top paddings for section-titles as when you click the point from the menu, it gets you redirected to the proper section, but there is no padding on top.


aliceblunt8 (changes requested):

- Add a favicon
- Location-related addresses / links / images should open google maps in a new tab target="_blank"
- Apply :hover effect for images on page 
- let's fix this when menu is open & user starts to scroll
Screenshot 2022-06-18 at 14 23 52

fenderukr:
1d8f23867e237982ebaf6e06c2fc602d6e614e63
plz check this commit

ArturErmolenko (changes requested):
- Было бы хорошо добавить сюда hover еффекты ,например менять цвет ссылок при наведении
Screenshot 2022-06-23 at 13 12 15

- При увеличение картинка перекрывает контент, так не должно быть
Screenshot 2022-06-23 at 13 12 32

PR #1229, opened 2022-06-28
miami

yura-atamaniuk:
https://yura-atamaniuk.github.io/layout_miami/


aliceblunt8 (changes requested):
- Location-related addresses / links / images should open google maps in a new tab target="_blank"
- this happens when menu is open and I scroll, needs to be fixed with z-index
Screenshot 2022-06-30 at 15 07 59

aliceblunt8 (approved):
good job!

also you can disable scroll when menu is open (see 18 point)

PR #1250, opened 2022-07-15
Task is implemented

ElinaBahirova:
https://elinabahirova.github.io/layout_miami/


ArturErmolenko (changes requested):
Good job 🔥 
Just fix form, form doesn't submit
1.Form shouldn't be submitted if some of the fields are not filled
2.Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)

marina-tilniak (changes requested):
ці 2 пункти меню не працюють
image

marina-tilniak (approved):

polosanya (commented):
1. I think it's better to make the main section take up the entire user's screen(100vh)
image
2. Fix logo link sizes
hover-link
3. Download an image from Figma without this thing
image
image
4. Placeholders move on hover
placeholders-hover
5. Background color doesn't match the design
image
6. Border radius of message input too
image
image

PR #1293, opened 2022-07-27
Landing

yungdena:
- [DEMO LINK](https://yungdena.github.io/layout_miami/)


AlinaMatuschak (changes requested):
https://www.loom.com/share/d4e47a88e979456aab917883e0e85251

AlinaMatuschak (commented):
- you can stop updating page with preventDefault
https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault
- and [reset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/reset) field after submit

src/index.html
```diff
@@ -1,13 +1,339 @@
...
+                  placeholder="E-mail"
+                  required
+                >
+                <textarea
+                  name="message"
+                  class="contact-us__field form-field form-field--textarea"
+                  placeholder="Message"
+                  required
+                ></textarea>
+                <button type="reset" class="button">
```
AlinaMatuschak, line 291: use js for resetting (go to qna if don’t understand how to implement it)

```suggestion
                <button type="submit" class="button">
```

AlinaMatuschak (approved):
Nice 👍

- can you vertical align your icons?
Screenshot 2022-08-03 at 20 56 42

- add max-width for header content
Screenshot 2022-08-03 at 20 58 17

PR #1349, opened 2022-08-04
1

vitaliy-beketov:
[DEMO LINK](https://vitaliy-beketov.github.io/layout_miami/)


src/index.html
```diff
@@ -1,13 +1,252 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" class="page">
```
Virass, line 2: We have only one html tag on the website, you can style him only.

src/index.html
```diff
@@ -1,13 +1,252 @@
...
+            <a href="#menu" class="icon icon--menu-open "></a>
+          </div>
+          <h1 class="header__title">
+            Take the Streets
+          </h1>
+        </div>
+      </div>
+    </header>
+    <nav class="page__menu menu" id="menu">
+      <div class="conteiner">
```
Virass, line 39: I think this class needs to name - container. Not conteiner

Virass (changes requested):
Looks good, good BEM, but I have several points. 
I think this landing can be looks good more if you add several animations on the page and respond to elements when your interact on the page.
For example, when you focused image in a gallery, I expect hover animation and zoom image. But zoom need realized without scale image block, only  scale image inside a block. 
`overflow: hidden`

src/index.html
vitaliy-beketov: it's not at all clear what's wrong

Virass: This is not good practice, for example for html 4 this syntax is not valid. For HTML 5 is a valid case, but theoretically, if somebody uses old browser they will see the error.
In principle, this will not be the biggest problem for the use / old browser, but nevertheless.

Virass (approved):

PR #1377, opened 2022-08-18
First try

Proph7000:
[DEMO LINK](https://proph7000.github.io/layout_miami/)


liliia-livandovskaya (changes requested):
1. Make a hover effect for this block
Знімок екрана 2022-08-25 о 13 45 03
 2. The speed of animations is the same throughout the landing page (for example, increasing when hovering or moving blocks when scrolling)

3. Make the hover effect there more user friendly
Знімок екрана 2022-08-25 о 13 44 20

4. Add favicon

stanislav-topikha (changes requested):
change your hover effect to a simple scale, maybe with a shadow
this one is too much)
Screenshot 2022-08-30 at 09 51 19

Form should be cleared after the submit

SanyaBratashchuk (changes requested):
1. Form: make fields required; add hover effects on it; form should reset after submittting;

liliia-livandovskaya (changes requested):
Great job! Add transition duration the same for each change on page
Знімок екрана 2022-09-06 о 17 32 46

liliia-livandovskaya (approved):
Great job! 👍

PR #1390, opened 2022-08-22
Develop

Hlazman:
DEMO LINK
https://hlazman.github.io/layout_miami/


liliia-livandovskaya (changes requested):
1. After clicking on the logo there is a 404 error. Fix it 
Знімок екрана 2022-08-26 о 11 37 57
Знімок екрана 2022-08-26 о 11 38 10

2. Make the shown on the hover image smaller. It should not hover over other images
Знімок екрана 2022-08-26 о 11 36 30

3. Add the transition duration so all the animations would change slower

stanislav-topikha (changes requested):
fix that white corners
Screenshot 2022-08-30 at 15 56 51

Better to make it the same as on the mock, not every recruiter gonna appreciate it)
Screenshot 2022-08-30 at 15 59 45

when you click on the logo in the footer there should be a smooth scroll to the top of the page

stanislav-topikha (approved):
Rejected by mistake

Hlazman:
> GJ 👍🏻

thanks a lot)

PR #1407, opened 2022-08-27
Develop

illia-sosnytskyi:
- [DEMO LINK](https://illia-sosnytskyi.github.io/layout_miami/)


Apollonia5:
Коли намагаюся заповнити форму поля очущуються і без натискання Send

OleksiiBlezniuk (changes requested):
GJ! There are few small comments to fix:

Please double check design, there are two fonts that used on mock, but there is only one on your page

I think `Explore` link should redirect somewhere to Compare bikes or Contact sections, it's strange that it redirects to the top of the page, but it's not critical

Let's add some hover effects to links in Contacts block (color change, underline, or smth like that)

OleksiiBlezniuk (approved):
Use `Inter` font for this heading
Screenshot 2022-09-06 at 12 39 01
Also create more meaningful messages to commits, it's hard to understand what have you changed from `fixed issues`

PR #1410, opened 2022-08-28
Develop

dmytristepanov:
- [DEMO LINK](https://dmytristepanov.github.io/layout_miami/)


anastasiia-nahaiska (commented):
GJ!
I noticed a few problems:
1. When I click on About Us, the page doesn't scroll to the About Us section.
2. The mentor told me that the phone in the menu should also have a color effect when it hovers. This isn't in the design, but it is a rule that all links should have a hover effect (You can choose the color from the design)
3. The main font family on this page is Roboto but in inputs have Arial ff. I think Misha is wrong in the video and inputs don't inherit ff.
image

anastasiia-nahaiska (approved):

src/index.html
```diff
@@ -1,13 +1,496 @@
...
+      <div class="header__container">
+        <div class="header__content">
+          <div class="header__top">
+            <a
+              href="#menu"
+              class="icons icons--menu"
+            >
+            </a>
+
+            <a href="/index.html" class="logo">
```
OleksiiBlezniuk, line 27:
```suggestion
            <a href="#" class="logo">
```

OleksiiBlezniuk: `#` is better, because now your page is reloading but it shouldn't

OleksiiBlezniuk (changes requested):
GJ! There are few small comments to fix
Links in testimonials block shouldn't refresh page, better to redirect somewhere to this people social media or just make text from them.
Same with links in logo, please check all links, and if you have `/index.html` in href, replace it with '#`
Also this href can be cause of this bug

https://user-images.githubusercontent.com/84275496/188460439-a9ee98ef-18e0-4451-9432-6bb9396dcf96.mov

Also lets add hover effects to links (color change/ scale/ underline/ shadow, etc.)

OleksiiBlezniuk (changes requested):
Add some hover effects to links in `about us`, `gallery` and `testimonials` blocks
Also `home` link in menu is not working

OleksiiBlezniuk (approved):
GJ! Nice hover effect on links 👍

PR #1411, opened 2022-08-28
Develop My Bike Landing Page

constantweb:
[DEMO LINK](https://constantweb.github.io/layout_miami/)


src/styles/utils/_vars.scss
```diff
@@ -1 +1,5 @@
-$c-gray: #eee;
+$page-color-first: #1d1d1d;
```
qoa11a, line 1: It will be not clear what color represent variables with names like first, second, third etc. You should use primary, secondary, accent, warning, error etc. Than there's sense in variable. Actually, there's a lot of naming conventions
https://css-tricks.com/what-do-you-name-color-variables/

src/styles/utils/_mixins.scss
```diff
@@ -1,6 +1,15 @@
...
+  transition: all 0.5s;
+}
+
+@mixin onTablet {
+  @media (min-width: 744px) {
+    @content;
+  }
+}
+
+@mixin onDesctop {
```
qoa11a, line 11: Typo
Desctop => Desktop

qoa11a (changes requested):

qoa11a (approved):

PR #1412, opened 2022-08-28
Bose landing

ihorandrianov:
- [DEMO LINK](https://ihorandrianov.github.io/layout_miami/)


Anna-Borodaienko (commented):
1. You don't need phone icon in the slide-menu
2. You have #e5e5e5 background-color of all page, but design has #FFFFFF
3. Links in the slide-menu don't work. But there shouldn't be any error, change these links to the main-page
4. The same for all links on the page - user can't receive 404 page error 
5. Clicking on the phone icon should show phone number only on desktop, not on mobile or tablet
6. In Contact section button has be "Send", not "Submit"
7. Form shouldn't be submitted if some of the fields are not filled (user has to receive messages, what field is not filled)
8. All fields should become clear after submitting
9. Section "Why buy direct from Bose" has to have 2 columns on Tablet (width 744px), but it has 1 column

Anna-Tkachenko (changes requested):
Good job!
Some comments:
1) When I click on some link in the opened menu it should redirect me to some block on the page, now it doesn't work, fix it
2) These links redirects to 404 error page, fix it
image
3) Add cursor:pointer on hover state for all interactive elements
4) Check the correct width for larger images
image

Anna-Tkachenko (changes requested):
Good job!
Small fixes:
1) Add cursor: pointer on hover state for all interactive elements (for example, opened menu, buttons)
2) Links in `Browse Bose products by category` section redirects to 404 error page - fix it

Anna-Tkachenko (approved):
Good job!

PR #1416, opened 2022-08-29
Bose audio devices landing page

TheVovchik:
- [DEMO LINK](https://TheVovchik.github.io/layout_miami/)


src/styles/blocks/icon.scss
```diff
@@ -0,0 +1,68 @@
...
+  &__tooltip {
+    position: absolute;
+    transform: translateX(-100%);
+    padding-top: $headers-svg-big / 4;
+    padding-right: 8px;
+    letter-spacing: 2px;
+    color: $page-color;
+    font-weight: 700;
+    opacity: 0;
+    pointer-events: none;
```
imaksweb, line 66: It would be great to add some smoothness when you show a tooltip

src/scripts/main.js
```diff
@@ -1 +1,22 @@
 'use strict';
+
+const formInputs = document.querySelectorAll('.form__input');
+const textArea = document.querySelector('.form__textarea');
```
imaksweb, line 4: Maybe the better option would be to add the same class to textarea as inputs have

src/index.html
```diff
@@ -1,13 +1,403 @@
...
+          >
+            <label>
+              <input
+                type="text"
+                name="name"
+                placeholder="Name"
+                class="form__input"
+                required
+              >
+            </label>
```
imaksweb, line 322: Is it necessary to wrap inputs by labels?

src/index.html
```diff
@@ -1,13 +1,403 @@
...
+            </a>
+            <a href="#item1" class="recommended__product-name">
+              Bose portable Smart speaker
+            </a>
+            <a href="#smart-home" class="recommended__product-category">
+              Smart home
+            </a>
+            <p class="recommended__buy-price">
+              $ 399.00
+            </p>
```
imaksweb, line 131: As option you can add "$" and ".00" via before and after pseudo elements, cause when you get the price from server you get only number and then you style it according to mockup

imaksweb (changes requested):
GJ! Just some minor things to fix if you think it's necessary

src/index.html
TheVovchik: image

Input with id "null" has no associated label  input-req-label

imaksweb (approved):
GJ🔥

silchyn (changes requested):
- I can enter a word as a phone number in the form
- Contacts links do not change color on hover

TheVovchik:
> * I can enter a word as a phone number in the form
> * Contacts links do not change color on hover

1) okay I will add a pattern to check a phone number;
2) I add smooth text shadow and have cursor pointer, I know that in checklist clause #4 we need to change color, I find text shadow much smoooother.

silchyn (changes requested):
Good job, but that is too much for a placeholder. You could have 'Number' as a label and an example as a placeholder, but not in that case. I would suggest leaving the placeholder a 'Phone number' and changing data validation to digits-only (type="number")

TheVovchik:
> Good job, but that is too much for a placeholder. You could have 'Number' as a label and an example as a placeholder, but not in that case. I would suggest leaving the placeholder a 'Phone number' and changing data validation to digits-only (type="number")

I will replace pattern that you could type with + or no and change standart alert message that will show what pattern you need to type.

silchyn (changes requested):
I think you are doing a lot more than needed. You could just set 'type' attribute to 'number' for phone number input

TheVovchik:
> I think you are doing a lot more than needed. You could just set 'type' attribute to 'number' for phone number input

how can it check the pattern of 12 numbers?! https://www.w3schools.com/tags/att_input_type_number.asp

silchyn (approved):
Cool!

PR #1419, opened 2022-08-30
Develop

Yeliseiev:
- [DEMO LINK](https://Yeliseiev.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Distance between this blocks should be bigger
Screenshot 2022-09-06 at 13 55 57

`Inter` font is used here
Screenshot 2022-09-06 at 13 56 58

Font in textarea is default, also looks like buttons has default border and outline
Screenshot 2022-09-06 at 14 17 53

Otherwise everything good 👍 Nice effect with footer, awesome transition on menu button 🔥 
Good job!

OleksiiBlezniuk (approved):

PR #1421, opened 2022-08-30
Develop

rstarovoit:
- [DEMO LINK](https://rstarovoit.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
You forgot to reset `h1` default margins 

Link here in first block should redirect somewhere to map, over 50 ... link can be just a text. There is no need to redirect user to top of the page from every link. Same with links in Testimonials, redirect somewhere to social medias of this people or make just blue text from it

After form submit page was redirected to page with error, fix it

Also links in logo redirects to page with error, fix it

rstarovoit:
> go redirects to page with error, fix it

rstarovoit:
> You forgot to reset `h1` default margins
> 
> Link here in first block should redirect somewhere to map, over 50 ... link can be just a text. There is no need to redirect user to top of the page from every link. Same with links in Testimonials, redirect somewhere to social medias of this people or make just blue text from it
> 
> After form submit page was redirected to page with error, fix it
> 
> Also links in logo redirects to page with error, fix it

Thnks 4 feedback! All done as you commented)

OleksiiBlezniuk (commented):
Approve, but I recommend you to add some hover effects to all blue links (for example make them darker). Also better to scale image in Testimonials instead of rotating - it will be more smooth and nice.

OleksiiBlezniuk (approved):
Please check previous comment

PR #1424, opened 2022-08-31
Develop

Shepeha-Andriy:
- [DEMO LINK](https://Shepeha-Andriy.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
You forgot to reset default `h1` margin

Links in logos redirects to page with error, fix it

Link `here` in first block should redirect to map, no need to redirect to top of the page

Links in testimonials block should redirect somewhere to social media of this people or be just a text

Background color of page should be white

Font in buttons should be bolder

Too big distance between Phone and Email, text in placeholders should be uppercased, placeholder in textarea is missed
Screenshot 2022-09-05 at 17 48 29

Also add hover effects to links

OleksiiBlezniuk (changes requested):
Links in testimonials section should be opened in new tab

Better to change color to blue in contacts block (use color from mock)

Add hover effects to links in gallery, testimonials, about-us blocks

OleksiiBlezniuk (approved):
Well done!

PR #1437, opened 2022-09-02
animation for phone number was added

Maksym845:
[DEMO LINK](https://Maksym845.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Text in gallery block is links, add here links to maps
Screenshot 2022-09-06 at 14 37 15

Add scroll-behaviour rule, page should be smoothly scrolled when user opened anchor link

Contacts is links too (map, phone, email)

This cards is not aligned properly, check design
Screenshot 2022-09-06 at 14 41 58


Also add hover effects to the links and images

PR #1440, opened 2022-09-02
create BOSE website

AndriyZhabynets:
- [DEMO LINK](https://AndriyZhabynets.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Something strange happened here 😄  If it was supposed to happen, then better to avoid it, this project will be reviewed by recruiters and they won't understand that
https://user-images.githubusercontent.com/84275496/188624245-6daf39e0-e737-4d53-a234-43f363c10663.mov

this button is different from mockup on hover
Screenshot 2022-09-06 at 14 27 45

Also phone button should be a bit bigger and aligned, please check mockup

It's strange that all links in `Recommended` and `Browse...` sections redirected users to top of the page, better to add links to some external resources

Also all links to external resources should be opened in new tab
Let's make inputs required, it looks strange now


Looks like cards in `recommended` section on tablets is not centered, paddings on tablets looks to small and some of images is bigger than it should be


Mobile version is ok 👍

OleksiiBlezniuk (changes requested):
Now all links in `Recommended` and  `Browse...` sections redirect to page with error, it shouldn't be like that, add links to external resources of make just text form it.

Also add some hover effect to links, it's hard to understand where is link and where is text, since it's black (also it's a good practice to add hover effects to elements which user can interact with)

Don't forget to add favicon to the page

Andr1yK (changes requested):
Something went wrong :(
image
image

Andr1yK (approved):
Good job 🔥 

P.S. Probably, a little big paddings look more pretty, but up to u😅

PR #1449, opened 2022-09-02
PR bose v.1

korniienkokarina:
- [DEMO LINK](https://korniienkokarina.github.io/layout_miami/)


poznianski (commented):
It seems good overall but lacks hover effects on elements. Also, there is no forward from phone/address elements. I guess grid would be better in this section instead of flex.

image

silchyn (changes requested):
- Why are menu links so much below where they should be?
- Some images' white background overlaps text on hover on small devices

silchyn (changes requested):
Good job, now the same with menu

silchyn (changes requested):
image

silchyn:
Horizontal scrollbar

silchyn (changes requested):
image

naziamoff (changes requested):
Looks good but there are some minor issues:
- Horizontal scroll is still with us
- Looks like there's something wrong with the icon on the burger menu hover
Знімок екрана 2022-09-12 о 10 55 53

- Make the button "View all" (under the gallery) clickable

naziamoff (approved):
great 🚀 🔥

PR #1450, opened 2022-09-02
Mayami landing

vovchuk-nazar:
- [DEMO LINK](https://vovchuk-nazar.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Something wrong with header 🥲 

https://user-images.githubusercontent.com/84275496/188613477-24ab1651-7b47-4710-adca-af4648d47449.mov

Something wrong with 3rd card in testimonials block

User shouldn't be able to submit empty form, also submit shouldn't reload the page or redirect to page with error

Also add some hover effects to links and images

PR #1451, opened 2022-09-02
Bose Landing Page

Naveed-Shoukat:
[DEMO LINK](https://naveed-shoukat.github.io/layout_miami/)


kamal-tufekcic (approved):
Very nice, good job with the grid

VinnieJ-2k20 (changes requested):
Looks promising, but requires improvements:

Checklist item 4

- Fonts all over the page are different from the design
image
image


- Header image is squashed on higher resolutions. Logo and buttons should have same content width limit as the rest of the page:
image

- Footer is the same:
image

- And on smaller screens it has the same problem:
image


- When I click on the link in menu, there's some strange jump of the page
image

- When menu is open, page scroll should be disabled
image

- Tablet page layout is broken
image

- And mobile as well
image

VinnieJ-2k20 (changes requested):
- Checklist item 4 not fixed.

- Footer image not fixed on higher resolutions. Also, there seems to be horizontal scroll.
image

- Phone and menu icons need to be black, not white
image

- There is even bigger horizontal scroll on tablet and mobile versions
image
image

VinnieJ-2k20 (changes requested):
Still cannot approve 100%.

- Make hover color change for links smooth

- Adjust the position of header image on mobile screens to look more like design. Now it doesn't look right
image

- Avoid squashing gallery images on mobile
image

- Add cursor: pointer and smooth change to buttons. 
image

ihor-jpeg (changes requested):
Screenshot 2022-09-20 at 13 03 07

PR #1455, opened 2022-09-03
Develop

vallerii:
- [DEMO LINK](https://vallerii.github.io/layout_miami/)


silchyn (changes requested):
- Why 'letter-spacing: -1.5px;'?
- Looks like you download icons as png instead of svg
- image

silchyn (changes requested):
Check letter spacing for text in the 'Why buy direct from Bose' section. I recommend avoiding setting letter-spacing for html and set it only where needed

silchyn (changes requested):
Good job, but
Check how you download icons
image
and
image

vallerii:
@silchyn what do you mean by how i download icons?

silchyn:
> @silchyn what do you mean by how i download icons?

@vallerii looks like your icons have worse quality than they should

naziamoff (changes requested):
Good job 🔥 
But there are some issues to fix
- Better use styles from the Figma mockup
Знімок екрана 2022-09-12 о 12 12 25
- The page shouldn't scroll in the menu
Знімок екрана 2022-09-12 о 12 14 03
- Check hovers of titles in the gallery section. Looks like they affect the behavior of other elements
- There are different fonts in different fields of the form

naziamoff (changes requested):
Nice!
Some additional fixes:
- Fix hovers on product cards. The product card should not increase on hover, only its picture
- Looks like sth is wrong with the font in the message textarea
Знімок екрана 2022-09-12 о 18 46 09

naziamoff (changes requested):
- Fix clickable area in the titles of products section. Looks like it is way too wide, should be only the text and probably some really small area around it

Знімок екрана 2022-09-13 о 09 52 50

naziamoff (approved):
Grats 🚀 🔥

PR #1466, opened 2022-09-04
BOSE Landing [Miami]

DenysKleimenov:
- [DEMO LINK](https://DenysKleimenov.github.io/layout_miami/)


silchyn (changes requested):
- Implement ':hover' effect for images in the 'Recommended' section
- Header should not take 100% height on small devices
- Phone number in the header overlaps the logo on small devices, think about a solution

DenysKleimenov:
> * Header should not take 100% height on small devices

I don't think that it's the necessary rule for this task, because I've seen a few solutions that were approved and they were taking 100% of height on small devices. Plus, I didn't find this suggestion on a checklist. 

> * Phone number in the header overlaps the logo on small devices, think about a solution

I think we can just remove phone number for small devices because it takes place only due to the **hover** effect and as far as I'm concerned we use hovers only for computers or laptops. Let me know what you think.

silchyn:
Good points :)

silchyn (approved):
Cool!

PR #1468, opened 2022-09-04
Miami-done

VovaDzhuryk:
- [DEMO LINK](https://VovaDzhuryk.github.io/layout_miami/)


anastasiia-nahaiska (commented):
Looks good! GJ!
I found a few problems:
Links on the contact us section and the phone number in the menu haven't hover effect. I know it's not in the design, but the mentor told me that we must do it. (You can choose some color from the design)
Also, the font-family in the form isn't the same. I think Misha was wrong and input doesn't inherit font-family
image

VovaDzhuryk:
> Looks good! GJ! I found a few problems: Links on the contact us section and the phone number in the menu haven't hover effect. I know it's not in the design, but the mentor told me that we must do it. (You can choose some color from the design) Also, the font-family in the form isn't the same. I think Misha was wrong and input doesn't inherit font-family image

VovaDzhuryk:
I made it according to the design, I think there is no need to change it!

PR #1469, opened 2022-09-04
Develop

andrew-satyr:
https://andrew-satyr.github.io/layout_miami/


OleksiiBlezniuk (changes requested):
Menu icons smaller than in mock, logo is too big, you forgot to reset default margins for `h1`, header should have 100vh height
Screenshot 2022-09-05 at 17 33 00


reset link styles, also this link shouldn't redirect to top of page, `here` should redirect somewhere to Miami map
Screenshot 2022-09-05 at 17 37 32

Links to external resources should open in new tabs
Links in Testimonials block should redirect to social media of this people or it should be just a text, no need to redirect user to top of the page

Now I can submit empty form, make fields required, also page shouldn't reload after form submit

Links in logo redirect to page with error, fix it

Also add some hover effects to links and images

Text style in gallery is different on mock on mobile devices (it's just a line)

Andr1yK (changes requested):
image

Disable resizing of textarea (or add only vertical resize and add min height)

Andr1yK (commented):
image

Very large images on tablet :(

naziamoff (changes requested):
- add favicon
- logo in your footer should redirect to the main page, but it redirects me to non-existing page
- you should not increase the whole card on hover, better do it only for the picture
Screenshot 2022-09-28 at 18 07 11

naziamoff (approved):
Well done 🔥 
I approve, but you should fix my comment

- I don't those should be links. Better make it just text
Screenshot 2022-09-29 at 15 17 33

Screenshot 2022-09-29 at 15 21 01

PR #1470, opened 2022-09-05
complete the task

barsuchellos:
- [DEMO LINK](https://barsuchellos.github.io/layout_miami/)


src/styles/blocks/button.scss
```diff
@@ -0,0 +1,38 @@
...
+    background-color: #005ecc;
+    border-color: #005ecc;
+    color: #fff;
+  }
+
+  &:hover {
+    color: #fff;
+    border-color: #005ecc;
+    background-color: #005ecc;
+    box-shadow: 0 13px 27px rgba(0, 83, 181, 0.46);
```
vaadyak, line 30: use variables for this shadow because u used it more than 1 time

vaadyak (changes requested):
User shouldn't be able to submit empty form
Form submit shouldn't reload page

OleksiiBlezniuk (changes requested):
You forgot to reset `h1` default margins, distance to the bottom of the page is too big

Add favicon

You forgot to add phone number to href
Screenshot 2022-09-20 at 18 27 52

Form shouldn't be submitted when it's empty, also it's submit shouldn't reload page
Screenshot 2022-09-20 at 18 30 19



Add transition to phone hover effect

Add hover effects to all links

barsuchellos:
@OleksiiBlezniuk thanks for review ,  I have already done transition but have some questions : 
at first , can you clarify where I should to fix bottom of the page (what the elements?). Secondly , where I must to add transition , bacause I have transition in my .place__photo-cotainer ?
image

Andr1yK (changes requested):
https://www.loom.com/share/7fbbbdbfd77442d380c51971d65e2d24

Andr1yK (approved):

PR #1472, opened 2022-09-05
Develop

tatyana22:
- [DEMO LINK](https://tatyana22.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Link `here` in first block should redirect somewhere to map, `over 50 ...` link can be just a text. There is no need to redirect user to top of page from every link. Same with links in Testimonials, redirect somewhere to social medias of this people or make just blue text from it 

Also add some hover effects to contact links

OleksiiBlezniuk (changes requested):
Color change looks good on contact links, let's remove underline, 'cause transition is not applied for it, so it looks bumpy.
Also let's apply some hover effect to the links in gallery and about us block, it's not clear if it's link or just blue text now (you can add animated line with pseudoelement, or add some shadow, or make it a bit more dark, don't forget about transition)

Would be nice if link in logo on the bottom of the page will redirect user to top of the page (replace `#home` with `#`)

`view all` button redirect to page with error

Looks like header is higher then 100vh 🙂 Fix it please

tatyana22:
> lery and about us block

ok!

OleksiiBlezniuk (approved):
Good job! Cool hover effect on links 👍 
Approve your job, but there is one issue that should be fixed - you forgot to reset default margin for `h1`, so distance between title and bottom of the header is too big (especially on mobile screens)

tatyana22:
> bottom of the header is too big (especially on mobile screens)

Thanks! I did reset.

PR #1473, opened 2022-09-05
Miami page

varkovroma:
[DEMO LINK](https://varkovroma.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Add favicon
You forgot to reset default `h1` margins (distance between title and bottom of the page is too big)
Background-color of the page should be white

User shouldn't be able to submit empty form (also submit of the form shouldn't reload the page)

Links in logo redirects to page with error

Logo in footer is too big

Don't forget to add hover effects to links and images

OleksiiBlezniuk (changes requested):
Form is not submitting now
You forgot to add phone number to phone icon link in header
When i disable menu - it redirects me to the page with error, fix it
Disable scroll when menu is open

Also don't forget to add hover effects to links 🙂

OleksiiBlezniuk (approved):
Good job!

PR #1476, opened 2022-09-05
Develop

vaadyak:
- [DEMO LINK](https://vaadyak.github.io/layout_miami/)

### DIFF
--- a/src/styles/blocks/button.scss
+++ b/src/styles/blocks/button.scss
@@ -18,7 +18,7 @@
   color: $colorBlue;
 
   cursor: pointer;
-  transition: all 0.3s;
+  transition: all 0.5s;
 
   &--primary {
     color: $colorWhite;
--- a/src/styles/blocks/page.scss
+++ b/src/styles/blocks/page.scss
@@ -33,7 +33,7 @@
     right: 0;
 
     transform: translateX(-100%);
-    transition: transform 0.3s;
+    transition: transform 0.5s;
 
     &:target {
       transform: translateX(0%);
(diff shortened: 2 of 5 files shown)


OleksiiBlezniuk (commented):
Please add `r` to this PR in spreadsheet when it will be ready for review

OleksiiBlezniuk (changes requested):
Please double-check design and checklist

You used wrong font almost everywhere
Background color should be white
Links shouldn't have horizontal paddings
Cards in testimonials is too big
User shouldn't be able to submit empty form
Form submit shouldn't reload page

OleksiiBlezniuk (changes requested):
Links in menu has horizontal margins, typo in `Contact` (it should be `Contacts`)
Screenshot 2022-09-07 at 11 48 09

Font is wrong, please check design

Page should be smoothly scrolled when we click on anchor links (use `scroll-behaviour`)

You forgot to reset `h1` default margins, distance between title and bottom of header is too big

Add hover effects to images and links

OleksiiBlezniuk (changes requested):
Line height is bigger here
Screenshot 2022-09-07 at 18 09 25

Link to mail is not working (use `mailto`)

Padding on mobile is too big
Screenshot 2022-09-07 at 18 10 39

Add hover effects to links in contacts, and also add transitions to hover effects, it shouldn't look so bumpy

OleksiiBlezniuk (commented):
Have you deployed your page after changes? I see noticed issues still

OleksiiBlezniuk (approved):
Good job! Would be nice to add hover effects to blue links (2,3 and 4) blocks

PR #1477, opened 2022-09-05
Develop

Veremchukk:
- [DEMO LINK](https://Veremchukk.github.io/layout_miami/)


Shepeha-Andriy (commented):
https://prnt.sc/gXToZWBQ5k7X     del scroll in menu
pictures in Gallery should increase on hover
https://prnt.sc/3C7DYqqs4mFO

OleksiiBlezniuk (changes requested):
You forgot to reset default title margins, distance between title and bottom of header is too big

Better to change color on hover in contacts block to blue (from mock) instead of grey

Form submit shouldn't reload page, also user shouldn't be able to submit empty form 

Add some hover effects to images and links

OleksiiBlezniuk (approved):
Good job!
Also I recommend to add hover effects to all links and to reset default input autocomplete styles (there is an information about it in checklist)

PR #1478, opened 2022-09-05
Develop

Arsen-Shpak:
[DEMO LINK](https://Arsen-Shpak.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Distance from logo to the top of the page is too big, also distance between heading and end of header is too big (you forgot to reset default margin)

Background of the page should be white


Distances between blocks is too big, check design

Icons on mobile verison is too big

Also add hover effects to links in gallery, testimonials and about-us block

Anna-Tkachenko (approved):
Good job! 🔥

PR #1481, opened 2022-09-05
bose landing created 75%

roman-zherebetskyi:
- [DEMO LINK](https://roman-zherebetskyi.github.io/layout_miami/)


roman-zherebetskyi:
finally... it works

OleksiiBlezniuk (changes requested):
Form submit redirects to page with error, fix it

Links on the page should redirect somewhere, now it looks like bug, best option is to redirect somewhere inside the page, also you can find some related to this product external resourses. If there are no resourses to redirect, just make text from this link. Probably text in `Browse Bose products` shouldn't be a link

Link in logo redirect to page with error

All links in menu should redirect somewhere inside the page, also

OleksiiBlezniuk (approved):
Good job!

PR #1499, opened 2022-09-08
create page Miami

rudenkorest:
- [DEMO LINK](https://rudenkorest.github.io/layout_miami/)


SanyaBratashchuk (changes requested):
1. It should be tel link too 
image
2. Add hover effects to images in gallery
3. Add hover effects to contacts
4.Page shouldn't refresh after form submit
5. Make fields required
6. Adress link should redirect to the google maps

PR #1503, opened 2022-09-08
Develop

bokorzita:
[DEMO LINK](https://bokorzita.github.io/layout_miami/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Miami landing page
 In this task, you will learn how to implement a landing page. To do that:
 
-- watch [the lesson videos](https://mate.academy/learn/css-advanced/css-advanced-landing#/video)
+- watch the lesson videos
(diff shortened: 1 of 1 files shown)


src/styles/blocks/header.scss
```diff
@@ -0,0 +1,114 @@
...
+    align-items: center;
+    flex-wrap: wrap;
+    list-style: none;
+    padding: 25px 20px;
+    position: relative;
+
+    @media(min-width: 744px) {
+      padding: 50px 72px;
+    }
+    @media(min-width: 1260px) {
```
amitpatiljc, line 63: these should be max-width and should be transfered to utility folders

amitpatiljc (changes requested):
Over all looks good, but needs lot of cleaning for example utilities like media query should be in extends.scss under utils folder
IMHO you can buid your page from here and push code in between for review
Watch videos as well so that you will get an idea

PR #1504, opened 2022-09-09
Miami landing

Oleh-Burii:


liliia-livandovskaya (changes requested):
Great job! Pay attention that the page should not be reloaded after the form submit. [checklist](Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)) p. 15  

Also, the speed of animations is the same throughout the landing page (for example, increasing when hovering or moving blocks when scrolling) [checklist](Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)) p. 12

Virass (approved):

PR #1507, opened 2022-09-10
MyBiKE

bradleymatjie:
- [DEMO LINK](https://bradleymatjie.github.io/layout_miami/)


kamal-tufekcic (changes requested):
The header image is a bit too short, it doesn't look good on FHD resolution, the nav menu animation is VERY slow, and the form/form buttons don't do anything, also, the images should have some animation/transformation, they're completely static.

VinnieJ-2k20 (changes requested):
Looks good, but needs some fixes.

Checklist item 4

- Logo and menu button on larger screens need to also have limited width, as the content
image

- Some alignment is off on larger screens
image

- Image is covering text here. Also, there's some strange button on the image
image

- Add `cursor: pointer` to buttons
image

- Form looks strange on tablets
image

- Very small paddings here on tablets
image

- Placeholder should start at the same level as the entered text
image

VinnieJ-2k20 (changes requested):
Step in the right direction, but not there yet.

- Content width should be limited, but not the picture:
image

- Hover effects here need to be smooth:
image

- This was not fixed. Make textarea same width as the inputs
image

VinnieJ-2k20 (approved):
Approved.

PR #1509, opened 2022-09-11
Created My Bike landing page

sapnachoudhary06:
- [DEMO LINK](https://sapnachoudhary06.github.io/layout_miami/)


kamal-tufekcic (changes requested):
Clicking on the X button leads to <https://sapnachoudhary06.github.io/> which doesn't exist, the form refreshes the page and it shouldn't, other than that everything looks spot on! Except maybe the MyBike title, which isn't positioned *exactly* like the design but IMO still looks good

Also, forgot to mention, the form fields should be required(?)

sapnachoudhary06:
@kamal-tufekcic I didn't understand your comment about My bike title... can you please point it out

sapnachoudhary06:
@kamal-tufekcic  Addressed your comments.

VinnieJ-2k20 (changes requested):
Looks very good, just couple of things to fix.

- Add smooth transition for hover styles on buttons:
image

- We don't need to show the number on small screens here:
image

VinnieJ-2k20 (approved):
Approved! Good job 👍

PR #1514, opened 2022-09-12
Develop

BZOzyurt:
- [DEMO LINK](https://BZOzyurt.github.io/layout_miami/)


kamal-tufekcic (approved):
Looks impeccable! Visually everything is spot on, so I'm giving it a pass, but you really should fix the X button in the nav menu before mentor review, if you press it it leads to https://bzozyurt.github.io/ which doesn't exist

VinnieJ-2k20 (changes requested):
Looks good, bot some items need to be fixed.

- Remove "Landing Page" from page title
image

- Menu open and close buttons should be on the same spot, so I don't need to move the cursor to press them. Also, yes, the link on the close button is broken and leads to 404.
image
image

- Distance here is net enough. Check the design
image

- When I hover over the button, it triggers hover over the image. Need to reset `padding: 100%` setting on larger screens.
image

- Add transition to hover styles of the button
image

- Header image on narrow screens looks broken
image

VinnieJ-2k20 (changes requested):
Almost done.

- Distance issue here was not fixed. Title should be higher regarding the header bottom line. Check the design
image

VinnieJ-2k20 (approved):
Good job 👍 
Approved.

PR #1525, opened 2022-09-18
Bose Landing Page

WebsterManyanga:
- [DEMO LINK](https://WebsterManyanga.github.io/layout_miami/)


kamal-tufekcic (approved):
Form submit is broken, it leads to a 405 error, and it also doesn't reset the form on submission, but other than that it looks good

ihor-jpeg (approved):

PR #1531, opened 2022-09-22
develop

NVlasenko:
[DEMO LINK](https://NVlasenko.github.io/layout_miami/)


SanyaBratashchuk (changes requested):
1. Use svg for icons
Ok, there is only header part, so keep working)

PR #1537, opened 2022-09-26
My bike mir

ruslik87:
[DEMO LINK](https://ruslik87.github.io/layout_miami/)


valerii-derkach (changes requested):
чудова робота!

давай зробимо, щоб форма при сабміті просто очищалась і не перезавантажувала сторінку

valerii-derkach (approved):

DarkMistyRoom (changes requested):
- let's make all form fields required

- I think would be better to improve the scale of these images.
image
here is an example. they scale with the container and don't overlap
https://www.loom.com/share/8b2dc0f6294d448e85b62b3ab2299c4c

- fix the background on the mobile. it looks different on the mockup
image

- check this gap
image

loralevitska (approved):
Good job!

PR #1547, opened 2022-09-28
added header 

zvir91:
[DEMO LINK](https://zvir91.github.io/layout_miami/)


src/index.html
```diff
@@ -1,13 +1,78 @@
...
+              class="icon icon--phone header__call"
+            ></a>
+
+            <a href="#" class="icon icon--menu">
+            </a>
+          </div>
+          
+          <h1 class="header__title">
+            <div class="header__rectangles">
+              <div class="header__rectangles-item header__rectangles-item--1 "></div>
```
TarasHoliuk, line 37: don't add each line separately
you can select them all end export from Figma as 1 SVG file:
image

src/styles/main.scss
```diff
@@ -2,6 +2,232 @@
...
+$headerTopPadding: 30px;
+
+.page {
+  font-family: "Inter", sans-serif;
+  &__body {
+    margin: 0;
+  }
+}
+
+.header {
```
TarasHoliuk, line 32: better split into separate files each BEM-blocks at the very beginning

src/styles/main.scss
```diff
@@ -2,6 +2,232 @@
...
+  align-items: center;
+  gap: 30px;
+  
+  @include onTablet {
+   --icon-width: 24px;
+   gap: 40px;
+  }
+}
+
+.menu {
```
TarasHoliuk, line 226: it should have `transform: translateX(-100%)` (shouldn't be visible) by default

TarasHoliuk (changes requested):
Please implement all other parts of the lending page
If you are stuck or have questions, ask them in chat or in the QnA session

stanislav-topikha (changes requested):
favicon
image

for some reason scroll doesn't work on initial load, until menu is opened

image transition should work both ways

page shouldn't reload on form submit, use` event.preventDefault()`

scroll to page sections should be smooth


Anastasiia-Svintsova (changes requested):
Please, check the figma mockup more carefully

Add transition here
Animation

Compare with the Figma design, please
image

CHECKLIST: Form shouldn't be submitted if some of the fields are not filled
Also add a transition to links

Animation

The location of the button does not match the design
image

The photo should look like this on mobile, follow the design
image

Anastasiia-Svintsova (changes requested):
1. Let's make the scale smaller so the pictures don't overlap each other
image
2. I can see some extra padding on a small screen and can't see the whole content in some places
image
image
3. I can see scroll when menu is open, but I shouldn't
image

PR #1549, opened 2022-09-28
added BOSE

Sasha-Green:
https://Sasha-Green.github.io/layout_miami/


SanyaBratashchuk (changes requested):
1. 404 after click on logo
2. Page shouldn't refresh after form submit
3. Add hover effects to contacts 
image

liliia-livandovskaya (approved):
Great job! 
1. The speed of animations is the same throughout the landing page (for example, increasing when hovering or moving blocks when scrolling)
Screenshot 2022-10-05 at 15 43 32

PR #1551, opened 2022-09-29
initial commit

NikolayGit1:
- [DEMO LINK](https://nikolaygit1.github.io/layout_miami/)


AlisaVlasova (changes requested):
1. these links should init telephone call
image
https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_link_phoneto

2. use svg format here 
image

3. let`s add some hover effects here 
image
image

4. check this gap, looks too small
image

5. form submission should not reload the page
6. form fields should be required
7. external links should open in a new tab 
image

PR #1561, opened 2022-10-03
Develop

Mikhaylo-Orlov:
[DEMO LINK](https://Mikhaylo-Orlov.github.io/layout_miami/)


TarasHoliuk:
This link leads to 404 error page
image

TarasHoliuk:
The page shouldn't reload after clicking `Send` button
image

vlad-demchuk (changes requested):
add transiton for menu button
image

add hover effects. users should know that's clickable
image
the same for these anchors. check all interactive elements. 
image

check the previous comment
The page shouldn't reload after clicking Send button

AlisaVlasova (approved):

PR #1571, opened 2022-10-06
Develop

Maks-Tsarenko:
[DEMO LINK](https://Maks-Tsarenko.github.io/layout_miami/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -289,7 +289,8 @@
               <form
                 class="contact-us__form contact-us__form--bottom"
                 method="get"
-                action="#">
+                action="#"
+                id="myform">
                 <label>
                   <input
                     name="name"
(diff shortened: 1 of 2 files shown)


alexandra-protyanova (changes requested):
Снимок экрана 2022-10-06 в 12 37 59
 I think it is not miami)

alexandra-protyanova (changes requested):
It is not necessary to ask for rearview until next section is ready) I am rejecting because of it is not the end of landing)

SanyaBratashchuk (changes requested):
1.Add hover effects for images
2. make fields required
3. page shouldn't refresh after form submit
4. should be a link to the google maps 
image
5. Change text color on hover
image

alexandra-protyanova (changes requested):
Page shouldn't be reload after form's submit, but should be clear

alexandra-protyanova (approved):
🔥

PR #1572, opened 2022-10-06
Bose

katya-suvorova:
https://katya-suvorova.github.io/layout_miami/


Anastasiia-Svintsova (changes requested):
Good job! Just fix a one small detail
The form shouldn't be submitted if some of the fields are not filled
Animation

polosanya (changes requested):
1. Burger menu icon doesn't match the markup
image
image
2. 404 error on logo click in burger menu
image
3. Phone number shouldn't appear on mobile on click
image

PR #1574, opened 2022-10-06
created MyBike

s3rhiienko:
https://s3rhiienko.github.io/layout_miami/


qoa11a (changes requested):
Great job! Just few comments:

- Form shouldn't reload the page after submission.
- Add cursor: pointer on all clickable elements (the button in the form).
- Why do you have cursor: pointer on the images if they're not interactive?
image

qoa11a (approved):
The form should be cleared after submission, it just shouldn't reload the page. You can use event.preventDefault() for it. I approve but please fix it.

PR #1583, opened 2022-10-11
Develop

Mykola-Busarhin:
[DEMO LINK](https://Mykola-Busarhin.github.io/layout_miami/)

I haven`t managed to solve the problem with the background image for header. What I mean is that it doesn`t adapt for different screen widths, though the content is adaptable. That`s why what I have is slightly different from the Figma mockup. Need your help.


SanyaBratashchuk (changes requested):
1. 404 after click on logo
Work is not completed, send on review only completed task if you have any questions feel free to ask in Slack

PR #1587, opened 2022-10-12
header, footer, 1st section added

andreyolin:
[DEMO LINK](https://andreyolin.github.io/layout_miami/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -252,7 +252,11 @@
 
           <div class="grid grid--tablet">
             <div class="grid__item grid__item--tablet--1-3 grid__item--1-5">
-              <form action="#" method="get">
+              <form
+                action="#"
+                method="get"
+                id="form"
+              >
                 <input
                   name="name"
                   type="text"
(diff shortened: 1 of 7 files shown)


mykola-koval (changes requested):
- checklist 1, 18, 15 etc
 go through the checklist once again
image
- address should open in a new tab
image
- use vector images for logos
image

mykola-koval (changes requested):
- burger menu looks different in the design
image

image

 - images should not cover each other on hover
image

- you should probably use a different approach for thisthe form submission (check out the checklist)
image

- add some transition duration on hover-effects (and i'd use a different more subdued color)
image

 - There should be no horizontal scrolling when hovering images on mobile

mykola-koval (approved):
gj overall 👍  but: 
 - let's use vector images for logos and icons
image
 - the button looks different in the design
image

 - add valadation to the form and clear it only if all the fields are valid

PR #1598, opened 2022-10-13
header

Khristina23042004:


mykola-koval (changes requested):
- CHECKLIST 13, 14, 15
 - ALWAYS use vector images for logos, icons etc
image

 - border radius on textarea is different in the design
image

 - button looks different in the design
image

image
image

Khristina23042004:
> 

what you mean about button styles isn't the same ?

mykola-koval (changes requested):
- this border on button is wrong
image
- form should clear on submit and there should not be any error
image

 - google maps should open in a new tab
image

mykola-koval (changes requested):
- email field in form should also be required (now it's not)
- this should not happen on hover 
image

 - images should not cover text below them on hover
image

PR #1601, opened 2022-10-13
MyBiKE

sh1gatsu:
DEMO LINK: https://sh1gatsu.github.io/layout_miami/


mykola-koval (changes requested):
image
 - if this text is a link, add some transition on it so I could know that these are links
image

mykola-koval (changes requested):
also fix the header on 320px 
image

PR #1606, opened 2022-10-14
header added

Yurii-Pozho:
[DEMO LINK](https://yurii-pozho.github.io/layout_miami/)


mykola-koval (changes requested):
- add validation on form
 - form should be cleared on submit
 - ALL external links (including maps) should open in a new tab

mykola-koval (approved):
gj 👍

PR #1609, opened 2022-10-14
miami

viktoriia-ovcharenko:
[DEMO LINK](https://viktoriia-ovcharenko.github.io/layout_miami/)


mykola-koval (changes requested):
- use vector images for logos
image
image
image
 - CHECKLIST 15
 - There should be no horizontal scrolling when hovering images on mobile
 - background color of the whole page is different in the design

mykola-koval (approved):
gj 👍 
image

PR #1614, opened 2022-10-15
BOSE header and footer were added

oksenyuk-valentyna:
DEMO LINK](https://oksenyuk-valentyna.github.io/layout_miami/


mykola-koval (changes requested):
looks like you haven't deployed, most blocks are missing and the page looks like this
image

mykola-koval (changes requested):
- there is a lot of differences with the design, for example let's have these backround image be full width
image

  - the submit form button looks different in the design

check out the design and fix your layout, then go through the checklist once again

mykola-koval (approved):
Good job 👍 , but there are still some improvements you can make:
 - when I click this button the scroll to top is not smooth
image

 - let's add some transitions to these links
image

image

 - all transitions on the page should have the same duration

PR #1615, opened 2022-10-15
all sections done

Anna-Shapovalova:
Live some comments, cause IDK why JS isn`t working in this case(


mykola-koval (changes requested):
- Should be no error when i submit the form. The form should be cleared and the page should not be reloaded.
image

mykola-koval (approved):
good job, congrats! 👍 🚀

PR #1619, opened 2022-10-16
Develop

alpopkov:
[DEMO LINK](https://alpopkov.github.io/layout_miami/


mykola-koval (changes requested):
- ALWAYS use vector images for logos
image

 - the icons should not crop, check all of them
image

denys-danyliuk (changes requested):
1 rule in checklist
2 rule, name it just Bose
add hover
Знімок екрана 2022-11-08 о 16 07 45


find some other color, let it be just a little bit brighter or darker than regular one
Знімок екрана 2022-11-08 о 16 06 27

I can't see icons on mobile
Знімок екрана 2022-11-08 о 16 11 24

also something wrong with this icon, check design
Знімок екрана 2022-11-08 о 16 11 29

PR #1621, opened 2022-10-17
implemented header block

Artem66:
[DEMO](https://artem66.github.io/layout_miami/)


mykola-koval (changes requested):
CHECK OUT THE CHECKLIST
 - page should not reload when we click on logos (use anchor links)
image

- this piece looks different in the design 
image

- there should be no errors when we click on logos
image

- pictures should not cover each others on hover
image

 - fix the mobile layout
image

 - you don't have any grid so the images in gallery won't stretch, let's implement it

mykola-koval (changes requested):
- ALWAYS use vector images for logos
image

 - there is a bigger gap between these links in the design
image

 - you don't have any grid implemented for the gallery, you should probably do this

illiarerun (approved):

PR #1622, opened 2022-10-17
header and footer are done (fixed)

shtht24:
[DEMO LINK](https://shtht24.github.io/layout_miami/)


mykola-koval (changes requested):
Looks like you haven't opened the checklist 
- your content on the page should be wider, check out the design

 - these images should not have any black background
image

 - none of the images should be stretched
image

 - the form looks different in the design 
 
 - all external pages (e.g maps) should open in a new tab
 
 - this button should not be black
image


Check the design again and fix all mistakes, and go through the checklist< it is mandatory

PR #1631, opened 2022-10-17
layout_miami MyBike v.1

Deemacr:
[DEMO LINK](https://deemacr.github.io/layout_miami/)


qoa11a (changes requested):
1. Inputs in form should be cleared after form submission (without page reload).
image
2. This button shouldn't redirect to 404.
3. Apply transition rules to whole class rather than to :hover. It makes transition works both ways and your animations looks smooth.
image
image
4. Make burger icon and cross icon be on the same level. It's better from UX point of view as you can click the same area to open/close menu. (On the design it's 36px while you have paddings of 54px in sum)
image

Deemacr:
Done

Anastasiia-Svintsova (approved):
It is better to reduce the scale so that the pictures do not overlap each other
image

PR #1633, opened 2022-10-17
MyBike-Header

92BS:
[DEMO LINK](https://92BS.github.io/layout_miami/)


qoa11a (changes requested):
Please finish the task before requesting code review.
You've probably forgot to deploy your changes.

polosanya (changes requested):
Good job! Looks really nice! Just found some moments that would be great to fix
1. Favicon is missing. 
image
2. There should be no scroll in burger menu
image
3. Location-related addresses / links should open google maps in a new tab. Also, it would be great to add some hover effect for all the links 
image
4. Add hover effects for images in The Details and Compare bikes sections
5. Wrong Indentations for tablets. Check yourself with the layout
image
6. Same problem with Compare bikes section
image
7. Burger menu color doesn't match the layout
image
8. Name and email placeholders have wrong font
image

Anastasiia-Svintsova (changes requested):
- The form has strange behavior. It should not be submitted if not all fields are filled in and it should also be cleared after submission. And the page should not scroll up.
Animation
- Add transition here 
Animation
- It doesn't look good. Let's highlight the text, not the block
Animation
- Doesn't match the design, check the letter spacing
image
- I can see the scroll when menu is open, but I shouldn't
Animation

polosanya (changes requested):
1. This hover effect looks a bit strange. Maybe let's try increasing the bike instead?
image
2. What is it? Can't see it on the markup
image
3. Text doesn't match the markup
image
image
4. Same here. Please check yourself everywhere
image
image
5. You shouldn't pass form data through URL. Prevent event or use method post for your form
image

polosanya (approved):
Well done! 🔥
Congrats, you did a great job!

PR #1634, opened 2022-10-17
layout_bose

MXMUK:


mykola-koval (changes requested):
gj, let's fix this 👇  

- images should not cover each other on hover
image

 - make this transition smooth

 - this should not happen at all
image

 - and this 
image

 - you should not have any horizontal scrolling on the page
 - implement grid for gallery

denys-danyliuk (changes requested):
this icon looks weird, also on small resolutions this icon and phone icon look small, they should be quite bigger
Знімок екрана 2022-10-24 о 17 03 56
Знімок екрана 2022-10-24 о 17 07 34

when I click on this button there are two ## added in the link
Знімок екрана 2022-10-24 о 17 05 12

denys-danyliuk (approved):

PR #1641, opened 2022-10-21
Develop

Michael1195:
- [DEMO LINK](https://Michael1195.github.io/layout_miami/)


loralevitska (approved):
looks amazing
it remains to correct a little

1. will be more effective to use `:hover` for your goods too
2. the photo in the header, text color (`Take the Streets` - dark color on a dark background is hard to read in the mobile version ) and background color on menu does not match the design
image

PR #1644, opened 2022-10-22
Bose landing

Oleksandr-Maslo:
https://oleksandr-maslo.github.io/layout_miami/


AlisaVlasova (changes requested):
1. add transition for smooth hovers
2. ipad
image
3. horisontal scrolll on mobile

PR #1645, opened 2022-10-23
Header was added

artem-hirzhev:
[DEMO LINK] (https://artem-hirzhev.github.io/layout_miami/)


stanislav-topikha (changes requested):
request review when the _whole_ page is done

PR #1648, opened 2022-10-23
Develop

MrFox32:
- [DEMO LINK](https://mrfox32.github.io/layout_miami/)


qoa11a (changes requested):
- User should be able to submit form

Anastasiia-Svintsova (changes requested):
Doesn't look good. I think form should be full width
image

Anastasiia-Svintsova (changes requested):
Please, fix the comment before requesting a review. If you have any questions feel free to ask them in the chat.

MrFox32:
I'm sorry, I didn't publish the latest version to gh pages
Ready now
ss

polosanya (changes requested):
Good job! Just found some moments that would be great to fix
1. It would be great to add a google maps link for the word `here`
image
2. IMHO, as a user, I expect that the picture will be opened in full-size on click, not navigate me to some page. So it's better to remove a link and cursor pointer for the image, or add the ability to open images in full-size. You may not fix it if you want
image
3. Hover effects here work strangely. Add them only for the icon, not the whole container
hover-miami

polosanya (approved):
Well done!

PR #1665, opened 2022-11-01
 layout_miami

Olha-Bychyk1:
[DEMO LINK](https://Olha-Bychyk1.github.io/layout_miami/)


Anastasiia-Svintsova (changes requested):
1. Add transition here
Animation
2. All links should open in a separate tab
Animation
3. Add hover effects here
image
4. Fix the form, please. The page shouldn't reload, and the form shouldn't be sent if some fields are empty
Animation

Anastasiia-Svintsova (changes requested):
- Transition should work both ways
Animation
- You still have the issue with the form. The page shouldn't reload
Animation

Bohdan-Maksymiuk (commented):
This PR is closed. Open it and request the review again, pls.
Ask in `fe_chat` if you have any questions

PR #1670, opened 2022-11-03
end block header

drhtka:
[DEMO LINK](https://drhtka.github.io/layout_miami/)


SanyaBratashchuk (changes requested):
Task is not ready for review. Re-request review when finish your task

PR #1676, opened 2022-11-03
Develop

Ukrainiane-panda:
[DEMO LINK](https://Ukrainiane-panda.github.io/layout_miami/)


SanyaBratashchuk (changes requested):
1. Click on logo shouldn' redirect from the site
2. Add hover effects for images
3. Make fields required 
4. Form submi tshoudn't trigger page refresh 
5. Disable scroll when menu is open 
image

SanyaBratashchuk (changes requested):
1. Add some hover effects for this links but not scale (color for example)
2. All external links should open in the separate tab

qoa11a (approved):

PR #1680, opened 2022-11-04
layout_miami_my_bike

Olha-Bychyk1:
- [DEMO LINK](https://<Olha-Bychyk1>.github.io/layout_miami/)


loralevitska (changes requested):
Submit the landing page for testing after all the blocks are ready, and before the review, check the demo link. It is not working now

loralevitska (changes requested):
request review after all the blocks are ready. Now you can convert your PR to draft 😉 
image

loralevitska (changes requested):
looks great!
let's improve it

1. I guess this text should be white 
image
2. click on [COMPARE BIKES](https://olha-bychyk1.github.io/layout_miami_my_bike/#compare-bikes) (menu) not scroll to title `COMPARE BIKES`

polosanya (changes requested):
Looks great! Just found a few moments that would be great to fix)
1. Phone number should appear on hover. Also, burger menu icon doesn't really match the design
image
image
2. Try to download icons as SVG, so they look okay even with zoom
image
3. What is it? Please remove it. Also, I thinks it's not a good idea to add cursor pointer on hovering images, as users would expect for some action to happen on click
image

polosanya (changes requested):
Almost perfect!
1. Icons should be bigger on tablet and desktop versions. 
Same with the cross in burger menu
Check yourself with the design.
image
image
2. Would be great to add some hover effect to the cross and these links
image
image

loralevitska (changes requested):
Almost done)
Check and fix indentation everywhere. For example remove double paddings

https://user-images.githubusercontent.com/70009205/211072268-d28e5b8e-4f01-4add-adb1-23aaa79e4c8e.mov

polosanya (changes requested):
Let's fix the desktop version, so it looks better. You can limit the content width if needed
1. Sizes don't match the design. Please check yourself with Figma
image
image
image
2. Page content should be narrowed. Icons should be bigger
image

PR #1690, opened 2022-11-07
THE MET

alina-isakova:
https://alina-isakova.github.io/layout_miami/


vlad-demchuk (changes requested):
Looks good!

add some hover effects and transition for the form fields
image
image
The logo leads to 404
logo

vlad-demchuk (changes requested):
Sorry, I missed one thing last time
I guess it's not `Miami`😊

Fix it and it will be done)

vlad-demchuk (approved):

PR #1698, opened 2022-11-10
header #1

Ssviatt:
[DEMO LINK](https://Ssviatt.github.io/layout_miami/)


DarkMistyRoom (changes requested):
- where is the hover effect for the phone?
image

- links in the menu don't work
- I can't scroll your page. I see only the header

SanyaBratashchuk (changes requested):
1. Add email validation
2. 405 after form submit 
3. form shoud reset after submit
4. Add hover effects to this links(but not a scale, color for example)
5. Add favicon

SanyaBratashchuk (changes requested):
1. Form shouldn't trigger page refresh

Bohdan-Maksymiuk (changes requested):
Screenshot 2022-11-17 at 14 25 49

* сheck font size here (mb smth else too) - looks not like on design
* form field should be validated and cleared after form submit

AlisaVlasova (changes requested):
1. form submission should not reload the page
2. check form horisontal spacing and width
image
3. check titles' font-size on desktop
image

SanyaBratashchuk (changes requested):
1. 
image
It doesn't look like in mockup

DarkMistyRoom (changes requested):
- Buttons should have 56px height
image
image

- Also check height of these images
image

- This is how should look the form. It looks wider on the mockup
image

AlisaVlasova (approved):

PR #1719, opened 2022-11-14
header added

lukacsja:
[DEMO LINK](https://lukacsja.github.io/layout_miami/)


naziamoff (changes requested):
Well done, keep going 🔥 🚀  

Lets add some slight hovers for the logo and images

meforoma (changes requested):
currently the form is kind of irresponsive - it just does nothing.

  event.preventDefault();
  form.reset();
});

lukacsja:
> 

thank you for providing the js. i think i managed to fix these

meforoma (approved):
👍

PR #1722, opened 2022-11-14
MyBike is in progress 

nurdanokcu:
DEMO LINK](https://nurdanokcu.github.io/layout_miami/


src/styles/blocks/header.scss
```diff
@@ -0,0 +1,51 @@
...
+
+    box-sizing: border-box;
+    height: 100vh;
+    padding: $headerTopPadding 0 88px;
+
+    @include onTablet {
+      padding: $headerTopPaddingTablet 0 130px;
+    }
+
+    @include onDesktop {
```
UmizDemud, line 27: This 204px added together with the element margin of ```h1``` makes the "Take the Streets" text appear in the middle. 

It should be closer to the bottom according to designs

src/styles/blocks/grid.scss
```diff
@@ -0,0 +1,35 @@
+.grid {
```
UmizDemud, line 1: I like this implementation

src/styles/blocks/button.scss
```diff
@@ -0,0 +1,43 @@
...
+  font-size: 16px;
+  line-height: 27px;
+  text-align: center;
+  color: $colorDark;
+  text-decoration: none;
+  border: none;
+
+  cursor: pointer;
+
+  &--send {
```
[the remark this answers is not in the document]
UmizDemud: You have auto margin and max-width, removing the max-width should do it.

src/styles/blocks/compare.scss
```diff
@@ -0,0 +1,47 @@
+.compare {
+  &__content {
+    display: grid;
+    justify-content: center;
+    grid-template-columns: minmax(auto, 392px);
```
UmizDemud, line 5: I generally don't use grid for mobile. 

But this code also works without any flaw.

src/styles/blocks/menu.scss
```diff
@@ -0,0 +1,94 @@
+.menu {
+  background-color: $colorGray;
```
UmizDemud, line 2: This isn't the same color with the design. But it looks better imo

UmizDemud (changes requested):
Took me a while to come up with anything wrong. Please do these corrections as you see fit. Once it is ready, I will take another look and approve.

src/styles/blocks/button.scss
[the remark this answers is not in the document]
nurdanokcu: I put this max-width for explore button. But it affected other buttons too. 
I added a modifier for explore button and moved max width and margin to that modifier. It should work now.

src/styles/blocks/compare.scss
nurdanokcu: I added this because bicycle pictures were getting so big right before hitting the tablet breakpoint. Then getting smaller again in tablet view.  And when I add max-width pictures weren't responsive anymore. I am open to suggestions if there is a better way to do it.

src/styles/blocks/menu.scss
nurdanokcu: Oh haha I picked the wrong color.  🤦🏻‍♀ Fixed it now.

src/styles/blocks/compare.scss
UmizDemud: No, I don't think anything wrong with this

nurdanokcu:
> Seems correct to me. Well done!

@UmizDemud Thank you.

naziamoff (changes requested):
- Let's add a hover effect for the logo
- There are different fonts inside the form
Screenshot 2022-11-18 at 11 36 27

src/scripts/main.js
```diff
@@ -1 +1,54 @@
...
+  }
+});
+
+const sendButton = document.querySelector('.button--send');
+
+sendButton.addEventListener('click', formSubmit, false);
+
+function formSubmit(event) {
+  event.preventDefault();
+}
```
meforoma, line 17: currently the form is kind of irresponsive - it just does nothing.
```

  event.preventDefault();
  form.reset();
});
```

meforoma (approved):
really nice 🚀 
love the halo-effect you applied for the gallery images on hover 👏

PR #1725, opened 2022-11-15
Lending mybike

MSA1991:
- [DEMO LINK](https://msa1991.github.io/layout_mybike/)


SanyaBratashchuk (changes requested):
1. Make transition work for both opacity and transform
image

SanyaBratashchuk (approved):

PR #1732, opened 2022-11-16
added header + menu

ihor-halkin:
[DEMO LINK](https://ihor-halkin.github.io/layout_miami/)


Anastasiia-Svintsova (changes requested):
Please, implement all sections before requesting the review

ihor-halkin:
> Please, implement all sections before requesting the review

Ok. In read.me file I found info about pull requests for each section separetly :)

polosanya (changes requested):
Looks great! Keep up the good work!
1. Scroll is disabled until user opens a burger menu. It's a critical bug that need to be fixed
scroll-bug
2. Let's try increasing just a bike on hover, not the whole container. 
I think it would look better. You don't have to fix it if you think that current effect is better.
image
3. What is it? Can't see it on the design
image
4. Form shouldn't pass the data through URL. Try preventing the event, or add method post to the form
image
5. Wrong email placeholder. There should be just 'Email'
image

loralevitska (changes requested):
your page is awesome) just little improvement

1. add paddings to your icons. Please fix all cases if present. It's currently hard to catch the clickable area
https://user-images.githubusercontent.com/70009205/203586855-3002dfc2-79f6-4f83-9290-f0e0c0126331.mov

tetiana-ivanchenko (approved):

PR #1735, opened 2022-11-16
created header, and menu

Lun4trick:
[DEMO LINK](https://lun4trick.github.io/layout_miami/)


src/index.html
```diff
@@ -1,13 +1,220 @@
...
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Cinzel:wght@400;700&display=swap" rel="stylesheet">
     <script src="scripts/main.js" defer></script>
   </head>
-  <body>
-    <h1>Miami</h1>
+  <body class="page__body">
+    <header class="header page__section">
+      <div class="container">
+        <div class="header__content">
```
ysrckr, line 16: if you can use the header__content as a container, if not rename container > header__container

src/index.html
```diff
@@ -1,13 +1,220 @@
...
+                class="logo__image"
+                alt="The MET"
+              >
+            </a>
+
+            <div class="header__icon-container">
+              <a href="tel:+1 212-535-7710" class="icon icon--phone"></a>
+
+              <a href="#menu" class="icon icon--menu-opener"></a>
+            </div>
```
ysrckr, line 30: with icon you are creating a new block so if you are using header__icon-container for margin and position purposes you should rename it icon header__icon and you can rename the icon to icon__link and so icon--phone to icon__link--phone but -- is a modifier it is correct to use icon__link--menu-opener but I'm not sure about icon__link--phone

src/index.html
```diff
@@ -1,13 +1,220 @@
...
+            Welcome
+            <br>
+            to The MET
+          </h1>
+        </div>
+      </div>
+    </header>
+
+    <nav class="page__menu menu" id="menu">
+      <div class="container">
```
ysrckr, line 42: Again you can rename it menu__container. Create a placeholder or a mixin for container and use it that way so all of the style is same between block__container

src/index.html
```diff
@@ -1,13 +1,220 @@
...
+                <span class="opening-hours opening-hours--hours">
+                  <div>10 am - 5 pm</div>
+                  <div>5 am - 10 pm</div>
+                  <div>closed</div>
+                </span>
+                <span class="opening-hours opening-hours--days">
+                  <div>Sun, Mon, Tue, Thu</div>
+                  <div>Fri, Sat</div>
+                  <div>Wen</div>
+                </span>
```
ysrckr, line 149: opening-hours is block its children should have opening-hours__something

src/index.html
```diff
@@ -1,13 +1,220 @@
...
+                <div>
+                  <div class="opening__hours">
```
ysrckr, line 151: I think you wrote it wrong. There isn't any block named 'opening'.  opening__hours makes it children of opening

ysrckr (commented):
In SCSS I didn't see break of BEM rules. They look correct in BEM perspective.

naziamoff (changes requested):
- the icons on the header actions look too big to me. Doublecheck on the mockup
- the scroll should be smooth when I click on one of the nav links
- some serious bug on the menu scrolling
- let's add slight hovers for the pictures and logo

meforoma (changes requested):
your entire page feels well built 👍 
The final part that's not fully as expected is the form, [checklist](https://github.com/mate-academy/layout_miami/blob/master/checklist.md) items follow:
form expectations

let's implement the above.

Lun4trick:
Thanks for suggestions guys! I think all what you asked for is implemented now!

PR #1757, opened 2022-11-17
My Bike landing page

oleh-dehtiarenko:
- [DEMO LINK](https://oleh-dehtiarenko.github.io/layout_miami/)


DarkMistyRoom (changes requested):
- logo shouldn't lead to 404
image

- no need to send a request on form submitting. Just reset fields and don't refresh the page

- add a hover effect to logo and it will be prettier to scale the whole container here, not only image inside
image

AlisaVlasova (approved):

PR #1768, opened 2022-11-18
Develop

grslckr:
[DEMO LINK](https://grslckr.github.io/layout_miami/)


meforoma (changes requested):
currently the form is kind of irresponsive - it just does nothing.

```

  event.preventDefault();
  form.reset();
});
```

grslckr:
> currently the form is kind of irresponsive - it just does nothing. Let's make it alive, so it at least resets and somehow react to submit.
> 
> ```
> const form = document.getElementById('form');
> //you'll need to provide your `<form` with `id='form'`
> 
> form.addEventListener('submit', function(event) {
>   event.preventDefault();
>   form.reset();
> });
> ```

the code you mention, it's already there

meforoma (approved):

PR #1776, opened 2022-11-19
miami_layout development

rmpolonskiy:


AlisaVlasova (changes requested):
re-request review when page will be ready

DarkMistyRoom (changes requested):
Good job, just fix these suggestions:
- Logo link to 404 error
- Address link should open on a new page
- All fields in the form should be required
- Page shouldn't reload on the form submitting
- And it will be prettier to add a hover effect to a container of these images
image
to make it works like here
image

SanyaBratashchuk (approved):

PR #1781, opened 2022-11-20
Add header and section about-us

yana-kotsulym:
https://yana-kotsulym.github.io/layout_miami/#


stanislav-topikha (changes requested):
request review when you are done with whole landing 
image


recheck the checklist

SanyaBratashchuk (changes requested):
1. Add hover effects for these links, but not scale (change color for example)
image

stanislav-topikha (changes requested):
404 error
image

same
image

DarkMistyRoom (changes requested):
Good Job, just a few suggestions:

- Logo has a too big clickable zone
image

- Will be great to add some hover effects to all links

DarkMistyRoom (approved):
I approve it but think about adding a hover effect to the logo

PR #1783, opened 2022-11-21
Bose landing

Rijo-J:
[DEMO LINK BOSE LANDING](https://rijo-j.github.io/layout_miami/)


meforoma (approved):
really nice implementation and own special touches
- the scale & twist of the images looks nice
- good catch on making not the 6-column, but 3-column grid - essential layout is the same but with less columns - that works totally fine if you provide own grid for each block
- and the layout maintains even at unrealistically narrow resolution - really great markup and styling 👍 
Monosnap BOSE® 2022-11-22 10-30-36

meforoma (approved):

PR #1787, opened 2022-11-21
Let the progress begin

IoghiStudio:
Just deployed , working on it.
[DEMO LINK](https://IoghiStudio.github.io/layout_miami/)


meforoma (changes requested):
@IoghiStudio so still one section since the commit 21 hours ago?

PR #1796, opened 2022-11-22
add project the met - header and menu sections

HelenShamshurina:
[DEMO LINK](https://HelenShamshurina.github.io/layout_miami/)


polosanya (changes requested):
Keep up the good work! 
Just found some moments that would be great to fix in this section
1. Let's make these phone numbers clickable, with opportunity to call
Don't forget to add some hover effects too
image
2. Also here both numbers navigate to the same phone number. Please fix it
image
image
3. Navigation should smoothly scroll user to the needed section, not just teleport
image

loralevitska (changes requested):
cool, those parts are fixed. Continue in the same direction 🦾

liliia-livandovskaya (approved):
Great job! 🔥

PR #1801, opened 2022-11-26
Header and footer

DmytroRozenberh:
[DEMO LINK](https://DmytroRozenberh.github.io/layout_miami/)


DarkMistyRoom (changes requested):
Please make the PR after finishing your work

SanyaBratashchuk (changes requested):
1. Add correct page title
2. Add favicon
3. 404 after click on logo
4. Add hover effects to the images 
5. Should be a link to the Google maps 
image
6. make all fields required
7. Form submit shouldn't trigger page refresh 
8. `object-fit: contain`
image

qoa11a (changes requested):
Form should be cleared after submit. Also it'd be nice to add basic HTML required rules.

DarkMistyRoom (changes requested):
- Will be enough to make the logo a link to your header, because it is what we expect from it
- These two links lead to the header. Make them lead to relevant sections
image

- Add transition to the hover effect on images to make it smooth like on the buttons
- According to the mockup content should be limited in width by 1020px
image

- Fix the background on the mobile screen
image

PR #1805, opened 2022-11-28
Landing page was created

Liliya-Kalinichenko:
[DEMO LINK](https://Liliya-Kalinichenko.github.io/layout_miami/)


polosanya (changes requested):
Keep up the good work! Some comments that would be great to fix:
1. There should be no errors on logo click (header, footer & burger menu)
image
2. Add some hover effect to the number
image
3. Add hover effects to the links or disable the cursor pointer for them.
`Over 50 years experience in love.` probably shouldn't be a link, but if you add the links, then make them do something or navigate to some external page, not just scroll to the top every time. 
image
4. Same here. Add hover effects to the names and make e.g. navigate to some social media, or make them just text.
5. What is this on the mobile version?
image

maxim2310 (approved):
Great work and separately excellent work with SCSS

Anastasiia-Svintsova (approved):

PR #1806, opened 2022-11-28
met done

Alessandranitti15:
[DEMO LINK](https://Alessandranitti15.github.io/layout_miami/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -335,7 +335,7 @@
 
             <div class="grid__item--8-12">
               <div class="contacts_info">
-                <form action="/">
+                <form action="#">
                   <input
                     class="form-field contacts__input"
                     name="name"
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,1 +1,9 @@
 'use strict';
+
+window.addEventListener('hashchange', () => {
+  if (window.location.hash === '#menu') {
+    document.body.classList.add('page__body--with-menu');
+  } else {
+    document.body.classList.remove('page__body--with-menu');
+  }
+});
--- a/src/styles/blocks/explore.scss
+++ b/src/styles/blocks/explore.scss
@@ -24,7 +24,6 @@
 
     &:hover {
       transform: scale(120%);
-      margin-bottom: 40px;
       z-index: 1;
     }
 
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -20,9 +20,6 @@
     margin: 0;
     background-color: #f8f8fa;
 
-    &--no-scroll {
-      overflow: hidden;
-    }
   }
 
   &__menu {
@@ -115,4 +112,8 @@
   @media (min-width: 1024px) {
     padding: $headerTopPaddingBig 120px;
   }
+}
+
+.page__body--with-menu {
+  overflow: hidden;
 }


amitpatiljc (changes requested):
Really amazing job ! Just some suggestions to make it more optimised please check this [video](https://www.loom.com/share/87bc350bd7404b4ba122955ab13ba3c9)

polosanya (changes requested):
Keep up the good work! Some comments that would be great to fix:
1. Number text doesn't match the design. 
image
Most likely that's the reason. Font-weight property should be set just with numeric value, without pixels.
https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#syntax
image
2. Add a hover effect to all links in the burger menu and line under the `Call to order`
image
image
3. Would be great to increase [z-index](https://developer.mozilla.org/en-US/docs/Web/CSS/z-index) for the image with hover
image
4. All the buttons should be interactive with some hover effect. Also, it doesn't match the design
image
image
5. Add hover effect to links
image
6. There is still an error on form submit(
image

polosanya (changes requested):
Hi, I see you haven't fixed all the previous comments.
Feel free to ask for help in the chat if you have any difficulties 🙂 

Also, found one more problem:
1. Hover effects work a bit laggy in this section. Make sure that you added them exactly to the images, not some container
met-hovers

Alessandranitti15:
Hi Oleksander, I actually fixed all you said.
1. font weight is correct.
2. hover on menu items is there
3. z index is done
4. I asked for help in chat and no solution was found, so maybe you can take a look at my code and help me
5. hover on contacts is there
6. in my server, there is no problem with the button "send", it send me on my homepage
Screenshot 2022-12-02 at 12 50 36
Screenshot 2022-12-02 at 12 49 55
Screenshot 2022-12-02 at 12 49 50

Alessandranitti15:
I have deployed the task with all the correction, but I need help with number 4

src/index.html
```diff
@@ -1,13 +1,376 @@
...
+                    <br>
+                    New York, NY 10040
+                  </a>
+                </div>
+              </div>
+            </div>
+
+            <div class="grid__item--8-12">
+              <div class="contacts_info">
+                <form action="/">
```
polosanya, line 338:
```suggestion
                <form action="#">
```
Try this. Also, would be great to use method post, to avoid passing the form data through URL
image

polosanya (changes requested):
A few more problems: 
1. There should be no scroll in burger menu
met-scroll
2. Try not to move text on image hover
met-image-hover
Also, please check my comment about the form.
What about button - make it navigate the user to some section (like you do in burger menu) or to some external page.

src/index.html
Alessandranitti15: Hi Oleksander, thank you for your helop. About the scrolling in the menu, I think I need JS to fix that, but I have *no idea* how to do it, because I haven't studied that yet. Can you please help me?

polosanya (changes requested):
Previous comments are not fixed.
If you have any questions, please ask them in the chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review 🙂

jstmpelowycz (approved):
nice job!

PR #1822, opened 2022-12-08
bose commit

jaxxxon78:
https://jaxxxon78.github.io/layout_miami/
для вирішення проблеми з placeholder в textarea


src/index.html
```diff
@@ -1,13 +1,379 @@
...
+                <textarea
+                  name="message"
+                  placeholder="Message"
+                  class="
+                    contact__field
+                    contact__field--last
+                    form-field
+                    form-field-textarea"
+                >
+                </textarea>
```
liliia-livandovskaya, line 330:
```suggestion
                <textarea
                  name="message"
                  placeholder="Message"
                  class="
                    contact__field
                    contact__field--last
                    form-field
                    form-field-textarea"
                ></textarea>
```

Virass:
Address link must be opened in a new tab
image

Virass (changes requested):
good job, but you need fix several points

src/index.html
```diff
@@ -1,13 +1,424 @@
...
+          <div class="grid grid--desktop">
+            <div class="grid__item grid__item--1-4">
+              <div class="recommend">
+                <div class="recommend__photo-container">
+                  <img
+                    src="images/photos/recommended1.jpg"
+                    alt="Bose portable Smart speaker"
+                    class="recommend__photo"
+                  >
+                  <div class="recommend__name">
```
Viktor-Kost, line 155: For the tablet screen text block should be centered

Знімок екрана 2022-12-14 о 19 13 03

src/index.html
```diff
@@ -1,13 +1,424 @@
...
+        class="page__section contact"
+        id="contact-us"
+      >
+        <div class="container">
+          <h2 class="page__section-title">
+            Contact us
+          </h2>
+          <div class="grid grid--desktop">
+            <div class="grid__item grid__item--1-5">
+              <form action="#">
```
Viktor-Kost, line 346: You need to set required fields so user will know either he is doing right or not

Знімок екрана 2022-12-14 о 19 14 58

src/styles/blocks/button.scss
```diff
@@ -0,0 +1,33 @@
...
+  justify-content: center;
+  background: #292929;
+  border-radius: 40px;
+  transition: all 0.2s;
+
+  &:hover {
+    font-weight: 700;
+    color: #292929;
+    background-color: transparent;
+    border: 3px solid #292929;
```
Viktor-Kost, line 21: Check your hover transition. It's not smooth because you add border for hover effect, but for static one you do not have border.
So there is a way how to fix it: use `box-shadow` instead of border
`box-shadow: inset 0px 0px 0px 3px #292929;`

Viktor-Kost (changes requested):

witflash (changes requested):
You didn't deploy your project, so it isn't available by the link in description.
Please, try to deploy it again.

AlisaVlasova (changes requested):
1. logo link leads to 404 page
2. prevent page scroll with opened menu 
image
3. form submit should not reload the page
4. fix textarea font 
image
5. set max width for content on large screens 
image
6. check header padding in the mockup 
image

UnknownAI (changes requested):
changing `div` to a `button` will allow seeing validation errors; 

to prevent reloading page, in main.js you need to add event listener for form submit event and use preventing default action function (checklist p15)

liliia-livandovskaya (changes requested):
Great job! 🔥 

There are points to fix:
1. There should not be errors when clicking on the logo 

Screenshot 2022-12-20 at 12 44 16

Screenshot 2022-12-20 at 12 48 17

Screenshot 2022-12-20 at 12 45 59

Screenshot 2022-12-20 at 12 48 17

2. You need to make the form inputs required. You can see an example [there](https://www.w3schools.com/tags/att_input_required.asp) 

Screenshot 2022-12-20 at 12 45 05

3. Remove this text from the placeholder. You can set the [validation](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#using_built-in_form_validation) 

Screenshot 2022-12-20 at 12 45 30

OlEzhA0 (changes requested):
the form should be cleared after submission
Screenshot 2022-12-22 at 3 05 33 PM

open and close menu on different distance from right side (1001px)
Screenshot 2022-12-22 at 3 08 16 PM
Screenshot 2022-12-22 at 3 08 29 PM


seems like should be on similar distance
Screenshot 2022-12-22 at 3 11 32 PM
Screenshot 2022-12-22 at 3 11 06 PM

if you have any questions just ping me in fe_chat

witflash (changes requested):
Hey! Almost done but here is the last problem.
You have used `button type="test"`. What is the reason for this?
You have to use `button type="submit"` to create the correct form.

AlisaVlasova (approved):

PR #1833, opened 2022-12-11
Develop

RomanHrytsiuk:
https://RomanHrytsiuk.github.io/layout_miami


DarkMistyRoom (changes requested):
Good. Please make the rerequest after finishing the whole landing

DarkMistyRoom (changes requested):
- it shouldn't link to the 404
image

- these links should call the phone client
image

- the bottom of the header looks cut
image

- images in this section looks squashed
image

- and make them zoom on hove with their containers, like [here]. And make the hover effect smooth in both directions (https://www.loom.com/share/8b2dc0f6294d448e85b62b3ab2299c4c)
- this button shouldn't have a permanent hovered state, it should be like all other buttons
image

- make all form fields required and the form shouldn't link to the 404 or reload the page, it should just reset on submitting
- fix the background on the tab 
image

- and on the mobile
image

SanyaBratashchuk (changes requested):
1. 404 after form submit
2. Add some hover effects for these links(e.g. Change color) 
image

SanyaBratashchuk (changes requested):
1. Align this icons
image

nazarmatsevych (approved):

PR #1851, opened 2022-12-17
Develop/Bike-page

OlhaYevstifieieva:
[DEMO LINK](https://OlhaYevstifieieva.github.io/layout_miami/)


src/index.html
```diff
@@ -1,13 +1,303 @@
...
     <script src="scripts/main.js" defer></script>
   </head>
-  <body>
-    <h1>Miami</h1>
+  <body class="page__body">
+    <header class="header">
+      <div class="container">
+        <div class="header__content">
+          <div class="header__top">
+            <a href="/" class="logo">
```
anna-inozemets, line 27: Maybe it is better to use class like header__logo logo (mix of classes), and add # in attribute href. (When we have just slash and click on logo we get "404 There isn't a GitHub Pages site here."

src/index.html
```diff
@@ -1,13 +1,303 @@
...
+              >
+              <h3 class="card__title">Agile ride 3</h3>
+              <p class="card__description">The lightweight frame that has earned its street tread as a sleek, agile ride.</p>
+              <span class="card__price">$ 2 090</span>
+            </div>
+          </div>
+        </div>
+      </section>
+
+      <section class="page__section" id="details">
```
anna-inozemets, line 142: U use just class page__section, I guess, it is sutible to add also class "details" - as it is main name of this block

src/index.html
```diff
@@ -1,13 +1,303 @@
...
+        <div class="container">
+          <h2 class="page__section-title">The details</h2>
+
+          <div class="details__card detail-card">
+            <div class="detail-card__photos grid">
+              <div class="grid__item detail-card__photo-container grid__item--tablet-1-4 grid__item--desktop-1-8">
+                <img
+                  src="./images/detail-1-1.jpg"
+                  alt="зображення кріплення для телефону"
```
anna-inozemets, line 151: You add alt on Ukranian, but we have only English language on the page, I think it is better to change alt on English

src/styles/blocks/about.scss
```diff
@@ -0,0 +1,43 @@
...
+  &__content {
+    @include onTablet {
+      display: grid;
+      grid-template-columns: repeat(6, 1fr);
+      column-gap: 24px;
+    }
+
+    @include onDesktop {
+      grid-template-columns: repeat(12, 1fr);
+      column-gap: 24px;
```
anna-inozemets, line 11: You have the same gap as onTablet, as onDesktop, I think you can use it just one time

src/styles/blocks/icon.scss
```diff
@@ -0,0 +1,63 @@
...
+
+    color: $main-accent;
+    font-size: 12px;
+    font-weight: 700;
+    line-height: 16px;
+    letter-spacing: 2px;
+    text-transform: uppercase;
+    opacity: 0;
+    pointer-events: none;
+    transition: opacity 0.2s linear;
```
anna-inozemets, line 32: I guess it is better to use same transition-duration value for all transitions on one web-site

anna-inozemets (approved):
Really cool landing page, I want to nuy a new bike after your work!🔥

src/styles/blocks/icon.scss
OlhaYevstifieieva: Yeas, I just forgot about this row )

Anna-Tkachenko (changes requested):
Good job! 
Some comments:
1) Check the correct position for this section on the large screen (probably you need to set some max-width for the content)
image
2) Will be better to make all field in form required
3) Check the position of background-image for the large screens
image
4) And also for the medium screens, probably you'll be able to find the way how to align it in more clear way
image

Anna-Tkachenko (changes requested):
Some comments:
1) Logo on the main page redirected to 404 error page, fix it
2) Check `tab` button with opened menu on the page, I can go to the other elements and I see them despite the opened menu, fix it
3) Check indents between sections on the large screens
image

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #1871, opened 2022-12-19
Develop

Anton-Zhytonbaiev:
[DEMO LINK](https://Anton-Zhytonbaiev.github.io/layout_miami/)


Avichez (changes requested):
Goood job 🔥 , but there is something that still needs to fix.
1) checklist 4, in the contacts section no animations on hover.
2) Please check the section's titles, there is something wrong with font-size, especially the title - "contact-us" compare it twice with the mockup.
image
3) When I'm changing elements by tab key the gallery is getting over the menu, try to add `z-index` for the menu.
image
4) And the last one thing, if I am not mistaken the text of the placeholder in the inputs has to be changing color during hover.

alexandra-protyanova (changes requested):
Снимок экрана 2022-12-27 в 14 55 35
On tablet image of bike should be fully visible.
Well done, just small fix left 🔥

alexandra-protyanova (approved):
Good job 🔥

PR #1873, opened 2022-12-19
Develop

NYasnaya:
https://NYasnaya.github.io/layout_miami/


vlad-demchuk (changes requested):
1. Better to move menu closer to the same place as menu opener
Dec-20-2022 11-47-31

2. add hovers to all clickable elements
image

3. form submission should not reload a page

DarkMistyRoom (changes requested):
- the logo from the menu links to 404
- the page shouldn't reload on form submitting
- make these images scale like on the video
image
https://www.loom.com/share/8b2dc0f6294d448e85b62b3ab2299c4c

nazarmatsevych (approved):
Great 👍

PR #1883, opened 2022-12-20
added header and menu

Vlad-Dihtiar:
[DEMO LINK](https://Vlad-Dihtiar.github.io/layout_miami/)


SanyaBratashchuk (changes requested):
1. 404 after click on logo 
2. Disable scroll when menu is open
Re-request review when finish whole landing

SanyaBratashchuk (changes requested):
Link is not working

PR #1886, opened 2022-12-20
the header and the menu were created

yaroslavna-horban:
[DEMO LINK](https://yaroslavna-horban.github.io/layout_miami/)


Anastasiia-Khudych (changes requested):
Good job! Nice menu button animation 👍 
But:
1) You forgot to add a favicon and title for your site. 
2) There is an error 404 that occurs on form submission. Forms should suggest what to enter.
3)  There shouldn't be any error when you click on the logo.
3) It seems, you didn’t reset default margins for the h1, h2, and h3 elements.
4) I think it will look better if you change the size of the image container on hover, but not only the image itself.
5) When I move through the menu using TAB, I can reach links in other sections.
6) Resizing the site I noticed, that your content touches the borders of the website, you should add padding to your container.
image
7) This section doesn’t look good on tablet-size screens.
image

ArturErmolenko (changes requested):
Good job 👍 

1. Header on tablet is different that on mockup
Screenshot 2022-12-26 at 13 35 35
Screenshot 2022-12-26 at 13 35 58

2. Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) make sure that there is no 404 error


https://user-images.githubusercontent.com/94292654/209544866-15991ca6-d235-4e5d-9f45-ba32d797554b.mov

Denchepornyuk (changes requested):
Screenshot 2022-12-27 at 16 04 55
1. remove horizontal resize from textarea


Screenshot 2022-12-27 at 16 05 37
2. inputs must be cleared after submitting the form

Denchepornyuk (approved):
good job 👍

PR #1893, opened 2022-12-20
header was added

ihor-soloviov:
[DEMO LINK] (https://ihor-soloviov.github.io/layout_miami/)


src/index.html
```diff
@@ -1,13 +1,339 @@
...
+                  required
+                >
+                <input
+                  name="email"
+                  type="email"
+                  class="contact-us__field form-field"
+                  placeholder="Email"
+                  required
+                >
+                <textarea name="message" class="contact-us__field form-field form-field--message"></textarea>
```
Avichez, line 297: Here is missing placeholder 'message' and the inputs using Arial font-family.
image

src/index.html
```diff
@@ -1,13 +1,339 @@
...
+
+                <button class="button">Send</button>
```
Avichez, line 299: Needs to add `border: none;`

Avichez (changes requested):
That's a great job 🔥 , especially animations on the first entry to the page.
1) Checklist 3, 6, in the deploy version I'm getting 404 error change in all Logos link to '#'.
2) For the menu needs to add `z-index` otherwise we will get this.
image
3) I'm not sure that you can change the colors of contacts, I understand that this is for animation, maybe to swap them?

src/index.html
ihor-soloviov: Yap. here I had strange but. I don`t know why but my placeholder was invisible, so I just cutted this part. I'll change it and pray for its vision

alexandra-protyanova (changes requested):
1. Checklist point 6
2. Let us follow the mockup, it is better to change color on hover)
Снимок экрана 2022-12-29 в 08 54 05

alexandra-protyanova:
In some resolutions, form & `Explore` button stretch too wide, it is not critical, but it is better to give them `max-width`

ihor-soloviov:
> 1. Checklist point 6
> 
>     2. Let us follow the mockup, it is better to change color on hover)
> 
> 
> Снимок экрана 2022-12-29 в 08 54 05

ArturErmolenko (approved):
Well done 👍

PR #1903, opened 2022-12-21
Develop

victorionov:
[DEMO LINK](https://victorionov.github.io/layout_miami/)


src/styles/header.scss
```diff
@@ -0,0 +1,127 @@
...
+        height: 24px;
+      }
+    }
+  }
+
+  &__menu-opener {
+    display: block;
+    width: 18px;
+    height: 12px;
+    background-size: cover;
```
nzrii-kysel, line 114: menu icon looks better with `contain`
```suggestion
```

src/styles/header.scss
```diff
@@ -0,0 +1,127 @@
...
+    background-position: center;
+    background-repeat: no-repeat;
```
nzrii-kysel, line 116: repeating styles
```suggestion
```

nzrii-kysel (changes requested):
nice work, few things that I've noticed:
- logos in menu and header blocks have different sizes
- clicking `tab` few times with menu opened will show you a little surprise
- first picture in Details section has weird `Discover the app` "button"
- links in Contacts section don't change color on hover (checklist point 4)
- menu is not scrollable on small screens (checklist point 17)
- there is an empty space below the footer

everything else looks great, gj

Anna-Tkachenko (changes requested):
Good job!
Some comments:
1) Logo in header redirects to 404 error page, fix it
2) Same with logo in the opened menu
3) Check correct indents here
image
4) Same here, check correct left / right indents for medium screens for all sections
image
5) Set some max size for the images, they're too big on medium screens
image

Denchepornyuk (approved):

PR #1906, opened 2022-12-22
created landing page Bose

MarinaOlenchenko:
[DEMO LINK](https://MarinaOlenchenko.github.io/layout_miami/)


vlad-demchuk (changes requested):
1. use svg for log and icons where it's possible for better quality
image
image

2. add hover to all clickable elements. users should know that it's interactive
image

3. make it smooth. add some transition to all hovers
image
image
image
and so on

4. form:
fields should be required
submission shouldn't reload page and clear fields afterwards
image

5. external links should be opened in a new tab
image

Check a checklist

SanyaBratashchuk (changes requested):
1. Check spacing here 
image
2. Add some hover effects for these links(change color for example)
image
3. Form shouldn't trigger form

SanyaBratashchuk (approved):
Add transition 
image

SanyaBratashchuk (approved):
There is no changes since last review. Add transition everything else looks good

PR #1913, opened 2022-12-24
new deploy after error

Vlad-Dihtiar:
[DEMO LINK](https://Vlad-Dihtiar.github.io/layout_miami/)


SanyaBratashchuk (changes requested):
1. Add some hover effects for these link (change color for example) 
image

SanyaBratashchuk (changes requested):
Form should reset on submit

SanyaBratashchuk (changes requested):
Message field should be required

DarkMistyRoom (changes requested):
almost there)
check the background on the mobile. the face shouldn't be cut
image

loralevitska (changes requested):
Please deploy your solution, and before review check demo_link. If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review
image

nazarmatsevych (approved):
👍

PR #1919, opened 2022-12-26
Bose landing develop

serbinm:
[DEMO LINK](https://serbinm.github.io/layout_Bose/)


src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,33 @@
...
+  @media (min-width: 1024px) {
+    @content;
+  }
+}
+
+@mixin onTablet {
+  @media (min-width: 600px) {
+    @content;
+  }
+}
```
nzrii-kysel, line 18: 600 is a bit too low for a tablet breakpoint

0xNazarii (changes requested):
well done, some comments:
- menu and header top paddings are different
- on form submit you should clear all input fields
- textarean should not be resizable
- horizontal scroll is visible starting from 1024px

src/styles/utils/mixins.scss
serbinm: - should not be pixel perfect tho
- made textarea resizable on purpose
- same with scroll
- I guess I'll pass on 600 width now

alexandra-protyanova (changes requested):
Fix @nzrii-kysel comments

PR #1922, opened 2022-12-27
[ADD] added readme.md and src/.

grabynsky:


witflash (changes requested):
I left some comments to improve your work.

1. You need to use `Inter` font for your landing page: 
image

2. Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)
image

3. Take attention to your link. Check if it works correct 
image

4. You have to add scroll for menu for small devices:
image
> Fix menu for small screens (if there is not enough space for all the menu items)

PR #1924, opened 2022-12-29
Develop

GUSILLUS:
[DEMO LINK](https://GUSILLUS.github.io/layout_miami/)


OlEzhA0 (changes requested):
1. form should clear after submission
2. add margin bottom to phone number
Screenshot 2022-12-29 at 3 18 25 PM
3. center your images
Screenshot 2022-12-29 at 3 20 32 PM

GUSILLUS:
@OlEzhA0 
If i add onclick evenet on submit
` <input type="submit" value="Submit" id="submit" onclick="submitForm">`

And script
`
function submitForm() {
    document.contact-form.submit();
    document.contact-form.reset();
  }`

Is that what i need to fix first task?

And do i need to delete `onclick="event.preventDefault()`?

Because previously mentor says that i need to add this to fix scrolling and reloading page after using "Submit button"

SanyaBratashchuk (changes requested):
Add hover effects for these links (change color e.g)

PR #1927, opened 2022-12-30
added all files for the landing task

alSemenyuk:
[DEMO LINK](https://alSemenyuk.github.io/layout_miami/)


src/styles/blocks/icon.scss
```diff
@@ -0,0 +1,48 @@
+.icon {
+  box-sizing: border-box;
+  display: block;
+  width: 16px;
+  height: 16px;
+  padding: 3px;
+  position: relative;
+  background-size: cover;
```
nzrii-kysel, line 8:
```suggestion
  background-size: contain;
```

nzrii-kysel (changes requested):
- checklist point 17
- form method should be `post` and make sure it doesn't reload the page on submit

apart from that, nicely done

ArturErmolenko (changes requested):
Good job 👍 🔥 

Just need to fix form :

- make sure that there is no 405 error

https://user-images.githubusercontent.com/94292654/211054846-24cb8d44-e8ce-483f-96cf-a4f437da90f8.mov

ArturErmolenko (approved):
Well done 🔥

PR #1931, opened 2022-12-31
add header

NikitaNevv:
https://NikitaNevv.github.io/the_met/


tetiana-ivanchenko (changes requested):
GJ, just few comments:
1. form submitting shouldn't redirect to another page
2. in checklist, point 4: Change text color on hover for phone, email and address
image
3. add transition on all hover effects
4. fix this block according to design
image

Bohdan-Maksymiuk (changes requested):
https://user-images.githubusercontent.com/94289038/211311213-2a6d67e2-5b39-45ae-a63a-1ed4c8d297fc.mov

* looks like your mobile page is broken

SanyaBratashchuk (changes requested):
Form shouldn't trigger page refresh

polosanya (changes requested):
Almost perfect!
1. Burger menu icon doesn't match the design
image
image
2. It would be great to reset the form on submit
image

mykhalenych (approved):

PR #1939, opened 2023-01-05
Develop

Vaskovics:
https://vaskovics.github.io/layout_miami/


DarkMistyRoom (changes requested):
check the font styles for the number
image

please make a request after finishing the whole page, thank you)

SanyaBratashchuk (changes requested):
Good job. Re-request review when you finish whole landing

Volodymyr-Melnyk-Fron10der (changes requested):
1. There is no image on design
image
2. Section recommended  should looks like on design
image
3. I think it's too much, scale it a little less + increase z index
image
4. Page is reloading on form submit, avoid it
5. decrease border radius 
image
6. There is no scrolling on small mobile
image
7. Smth wrong with image width on small height
image
8. Check all paddings on design
image
9. Check it on design as well
image

slavanosatskyi (changes requested):
1. You form is not cleared after submitting
2. Please, remove default outline for edit fields
Screenshot 2023-02-20 at 18 16 11
3. no hover effect for phone number (change text color)
Screenshot 2023-02-20 at 18 19 39
5. no hover effect for submit button. It should look like other buttons - black background and white on hover 
Screenshot 2023-02-20 at 18 21 16

Viktor-Kost (changes requested):
Hi!)
Almost done!)

1) add hover effects at menu
Screenshot 2023-02-23 at 14 22 27

2) don't fix container width for mobile screens. Just use padding (I set 20px as an example)
Screenshot 2023-02-23 at 14 25 35

SanyaBratashchuk (approved):

PR #1946, opened 2023-01-06
site about bikes was created

antonina-klishch:
https://antonina-klishch.github.io/layout_miami/


agniya-i (changes requested):
- I open and close the menu, then go to the section and click back using browser - it goes back to the open menu (which is fine) but the section images appear
Screenshot 2023-01-07 at 15 34 35

-  Add scale for sections images on hover 

- Make form fields require 
- Link must be open in a new tab

Screenshot 2023-01-07 at 15 35 04

- On the tablet, the image should be centered

Screenshot 2023-01-07 at 15 43 49

agniya-i (approved):
👍

PR #1953, opened 2023-01-09
created header, nav and footer. In future i'll add one block at a time

yurakosh701:
[MYBIKE landing](https://www.figma.com/file/NZQAIydtHo5QkINyGLHNcq/BIKE-New-Version?node-id=0%3A1)

[DEMO LINK](https://yurakosh701.github.io/layout_miami/)


loralevitska (changes requested):

1. Number should be center with icon 
image
2. When hovering over the phone icon, a number appears, but I cannot hover over this number, it disappears. Fix it
3. Menu background color doesn't match with design
image. Should be '#C2C2C2'

qoa11a (approved):

qoa11a (approved):
Good job!
Let's just clear form inputs after submit.

polosanya (changes requested):
1. Let's add a favicon for the page
image
2. Logo in burger menu has a bit worse resolution. Try to fix it
image
3. Let's add hover effects to all images on page
image
4. Same here. Also, try to download photo without this button. It's absent on the design
image
5. Add hover effects for these links. Also, Google maps link should be opened in a new tab
image
6. Burger menu icon doesn't match the design
image
image

qoa11a (approved):

PR #1963, opened 2023-01-11
Develop

MeryamWazeen:
[DEMO LINK](https://meryamwazeen.github.io/layout_miami/)


polosanya (changes requested):
Keep up the good work!
1. Try to fix all the problems in the code
image
2. Add a favicon and correct title to your page
image
3. There should be no error on logo click (both main page and burger menu)
image
4. Let's add increasing hover effect for all the images
image
5. Same with all interactive elements. Buttons should have cursor pointer and some effect on hover
image
6. Fix this problem
image
7. Add hover effects to the links. Also, address links should be clickable and open a Google Maps page in a new tab
image
8. Don't pass the form data through URL. Also, there should be no error on form submit
image

polosanya (changes requested):
Please fix all the previous comments)
There is still an error on logo click in burger menu
image
Not fixed
image
image
image

MeryamWazeen:
I don't know what's wrong with the button"view all exhibitions" !! in mobile view it works but in desktop it does not

polosanya (changes requested):
Good job!
But I see that problems 5, 7 and 8 from previous comments are still not fixed. Let's try to fix them all before moving forward)

MeryamWazeen:
> Good job! But I see that problems 5, 7 and 8 from previous comments are still not fixed. Let's try to fix them all before moving forward)

I could not fix problem 5 . can you plz help me with that?

polosanya (changes requested):
Please ask for help in Slack chat if you have any difficulties. It will be easier to communicate there than on Github
5. Add `cursor: pointer` to button and effects on hover, how you do it for images and links.
7. Add `target="_blank` attribute to google maps link in HTML, so it opens in a new tab
https://www.freecodecamp.org/news/how-to-open-a-link-in-a-new-tab/#:~:text=You%20can%20use%20the%20target,opening%20anchor%20tag%20like%20this.
8. Add method post to the form or prevent an event using js
https://www.w3schools.com/tags/att_form_method.asp

PR #1964, opened 2023-01-11
add header, footer

Anatolii-Polishchuk:
https://Anatolii-Polishchuk.github.io/layout_miami/


loralevitska (commented):
Please deploy your solution, and before review check demo_link
image

loralevitska (changes requested):
Please deploy your solution, and before review check demo_link
image

marina-tilniak (changes requested):
- logo shouldn't lead to 404 page 
image

- use svg for logo 

- animation isn't smooth 
- open in new tab
image

- after submitting form should be reset (write in chat if you have questions)
- add hover effect to images and links
- text here shouldn't fit whole width
image

vlad-demchuk (changes requested):
1. background doesn't follow the mockup
mockup
image

your implementation 
image

2. button should clear fields
image

nazarmatsevych (approved):

PR #1967, opened 2023-01-11
Header and Footer created

VKonoplianyi:
[DEMO LINK](https://VKonoplianyi.github.io/layout_miami/)


agniya-i (changes requested):
Good job! 
Pay attention that 404 error appears after click on logo

Re-request review when you finish whole landing

vlad-demchuk (changes requested):
after form submit
image

implement this checklist point
image

Volodymyr-Melnyk-Fron10der (changes requested):
Perfect landing!
But there is checkpoint like:
image

So, your menu is fixed, but isn`t scrollable if there is not enough space
image

DarkMistyRoom (approved):
almost there just add transition to these elements
image

VKonoplianyi:
> almost there just add transition to these elements image

Transition was almost there :)
image
image

PR #1968, opened 2023-01-12
Header&Menu added

Intellium7:


Virass (changes requested):
Complete work with landing and then response review

marina-tilniak (changes requested):
- open in new tab 
 image

- form should reset after submit(use js and ask in chat if have questions)

- would be good to add some hover effects to links

Intellium7:
> * form should reset after submit(use js and ask in chat if have questions)
   **added JS code even though I haven`t understood how it worked yet**

SanyaBratashchuk (changes requested):
1. All external links should be open in the separate tab 
2. Border-radius 
image
3. Make all fields required
4. Form submit shouldn't trigger redirect to other page

SanyaBratashchuk (changes requested):
Link is not working

Intellium7:
> Link is not working

fixed

Volodymyr-Melnyk-Fron10der (changes requested):
I don`t see previous comments were fixed(  Please pay more attention to it
+ my comments:
1. Menu is not scrollable when there is not enough space
image
2. Background image is not sized correctly (fix it to the right side, not center)
image
3. Links color is not changing on hover
image

Volodymyr-Melnyk-Fron10der (changes requested):
1. Avoid 405 error after submitting the form
image
2. Menu is not scrolling when there is not enough space
image

SanyaBratashchuk (changes requested):
This images doesn't look like in mockup 
image

SanyaBratashchuk (approved):
Write more meaningfull commit messages 
image

PR #1973, opened 2023-01-13
added layout of header

IamLoren:
https://iamloren.github.io/layout_miami/


IvanFesenko (changes requested):
On logo click got 404 error

SanyaBratashchuk (changes requested):
Good job. Keep working!

VitaliyBondarenko1982 (changes requested):
Good job.
Needs some improvements
- check backgoriund-position for header image
  on some width too hard to see phone numbers
  
Screenshot from 2023-03-12 16-48-17
- images overlap on hover
 suggest add overflow: hidden for image container
 
Screenshot from 2023-03-12 16-49-40

- after submit form fields should be cleared.
- add cursor pointer for send button
- check and add if ii is need transition for all elements which have hover effect (send, view all buttons)

etojeDenys (changes requested):
1. we don't need this box-shadow
image
2. on logo click page doesn't need to re-reload
3. Add a smooth scroll for the whole page

VitaliyBondarenko1982 (changes requested):
needs some little bug fix.
after submit leads to 405 page
for fix it add event.preventDefault() to listener
Screenshot from 2023-03-13 12-32-42
Screenshot from 2023-03-13 12-32-11

IvanFesenko (changes requested):
Disable scroll in case whrn menu is open
```
.page__body--with-menu {
  overflow: hidden;
}
window.addEventListener('hashchange', () => {
  if (window.location.hash === '#menu') {
    document.body.classList.add('page__body--with-menu');
  } else {
    document.body.classList.remove('page__body--with-menu');
  }
});


```

Check this case
https://user-images.githubusercontent.com/61904995/224726929-d361194d-ca08-4500-9cd3-a8bd10330569.mp4

2023-03-13_16h12_10

etojeDenys (changes requested):
1. let's fix this mismatch
image
2. and prevent page reloading on logo click

OleziO (approved):

PR #1976, opened 2023-01-14
MyBike.

Pavlo-Petrashevskyi:
[DEMO LINK](https://Pavlo-Petrashevskyi.github.io/layout_miami/)


DarkMistyRoom (changes requested):
Good! Please rerequest after finishing whole project

it also should be a tel link
image

Pavlo-Petrashevskyi:
> Good! Please rerequest after finishing whole project
> 
> it also should be a tel link image

vlad-demchuk (changes requested):
1. add some scale to images(checklist)
image

2. implements this checklist point
image

3. external links should be opened in a new tab
image

DarkMistyRoom (changes requested):
- make these images scale separately
image

- make the cursor a pointer over this button
image

- the form should reset on submitting
- remove the background while autocompliting
image

Volodymyr-Melnyk-Fron10der (changes requested):
Excellent landing! Only a few fixes to do:

1. Lines have to align to the right side
image
2. Implement this chekpoint
image
3. Email -> E-mail
image

Volodymyr-Melnyk-Fron10der (approved):
Great job! Approved!

Only one stuff I would change for future:
Don't use image to make this animation, create it from scratch with html + css and change `width `property - it will looks more smooth 
image

PR #1982, opened 2023-01-17
header menu sections finished

domkab:
[DEMO LINK] (https://domkab.github.io/layout_miami/)


polosanya (changes requested):
There should be no error on logo click
image
Same problem with logo and cross in burger menu
image

polosanya (changes requested):
Keep up the good work! You can re-request the review when other sections are ready
1. These icons should be bigger on desktop. Check yourself with the design
image
2. Background color is different
image
image

polosanya (changes requested):
Almost perfect! A few comments to fix:
1. There should be 40px between icons, not 24
image
image
2. Can you download an image from Figma without this thing?
image
3. Font in form placeholders doesn't match the design
image
image
4. Let's add hover effects to these links 
image
5. Address with two 's'. Also, make sure that this link is opened in a new tab
image

polosanya (changes requested):
Let's polish it up)
1. Try this to download an image without that thing
image
2. Address with two 'd' I meant)
image
3. Check if it's the right font from design
image
image

qoa11a (changes requested):
form should be cleared after submit

qoa11a (approved):

PR #1994, opened 2023-01-21
develop bike

Roman-Kulchytskyi:
https://roman-kulchytskyi.github.io/layout_miami/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -18,7 +18,7 @@
       href="./images/favicon.svg"
     >
   </head>
-  <body class="page__body page__body--with-menu">
+  <body class="page__body">
     <header class="header">
       <div class="container">
         <div class="header__content">
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -8,6 +8,12 @@ window.addEventListener('hashchange', () => {
   }
 });
 
+if (window.location.hash === '#menu') {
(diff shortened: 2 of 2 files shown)


mykhalenych (changes requested):
Nice work, but missed requirement:

Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)

polosanya (changes requested):
Looks great! Just found a few small problems:
1. There is an error on logo click in burger menu
image
2. Some placeholders have the wrong font
image
image
3. Google maps link should be opened in a new tab
image

witflash (changes requested):
Good job!
But you didn't fix one problem: 
> 3. Google maps link should be opened in a new tab

OlEzhA0 (changes requested):
check this bug
Screenshot 2023-01-25 at 10 56 50 AM

great job

OlEzhA0 (changes requested):
when I open your page first time, here is class `page with menu` and scroll unavailable 
could you check this issue please

https://user-images.githubusercontent.com/60295799/214805904-698f5946-d09f-470b-9a99-7cdbde765337.mov

vlad-demchuk (approved):

PR #1996, opened 2023-01-22
Develop

venher-v:
[DEMO LINK](https://venher-v.github.io/layout_miami/)


Volodymyr-Melnyk-Fron10der (changes requested):
1. Test is failed
image
2. Up button is not working
image
3. Implement this checkpoint
image
4. Add transition when phone numbers are appearing (keep in mind that transition should be consistent through the whole page)
image

SanyaBratashchuk (changes requested):
1. 404 after click on logo in menu
2. Change text color of these links on hover 
image

PR #1997, opened 2023-01-22
Header added

dissolveyo:
[DEMO LINK](https://dissolveyo.github.io/layout_miami/)


DarkMistyRoom (changes requested):
Good! Keep going and make request after finishing whole page

DarkMistyRoom (changes requested):
Very good, just a few comments:
- the cursor should become a pointer over buttons
image

- add hover effects to all links
- addresses should open a new tab
image

- all form fields should be required and the form submitting shouldn't reload the page

DarkMistyRoom (changes requested):
Almost there, just reset the form on submitting

slavanosatskyi (approved):
Nice work!

ivbutsykin (commented):
**Final review**

Perfect, I didn't find any issues. Add this work in your cv

PR #2010, opened 2023-01-27
adaptive header with navigation menu created

kirill-vypirovskyi:
https://kirill-vypirovskyi.github.io/layout_miami/


stanislavcodes (approved):
Excellent!

OleksiiBlezniuk (approved):
Good landing, nice success form message 🔥 There are few not critical issues, but better to fix them
Let's center this image on tablets, better to avoid cropping (especially if on design is centered and all content is visible)
Screenshot 2023-02-11 at 12 06 48

Button `explore` redirect user to the top of the page, and it's not logical, let's redirect to contacts form for example. Or to third party resource

PR #2025, opened 2023-01-30
Added a landing page

oleksusov:
[DEMO LINK](https://oleksusov.github.io/layout_miami/)


polosanya (changes requested):
1. Make sure that the transition applies to the default state of an element, not hover, so hover effects work smoothly both ways. Also, try to download the image without that button in the corner
hover-transition
2. Placeholders don't match the design
image
image
3. Add hover effects to these links
image

polosanya (changes requested):
Let's polish it up)
1. Phone number should appear on hover 
image
image
2. Background color doesn't match the design
image
image
3. Position of background image doesn't match the design
image
image

polosanya (approved):
Looks cool! You did a great job! 
Well done 🔥

PR #2036, opened 2023-02-03
Solution

Ihor-Zakharov:
[DEMO LINK](https://ihor-zakharov.github.io/layout_miami/)


ivbutsykin (changes requested):
Hey, looks good)

Keep doing, but notice: you should add a favicon and looks like space between phone and burger button should be more

slavanosatskyi (changes requested):
Nice work :) though some changes are required

1. On submit form the page is reloaded
2. Form shouldn't be with empty fields (all fields are required)
3. No animations of images on hover
4. Clicks on address should open in a new tab
5. address and phone links should change their color on hover

Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Only second comment was fixed from previous comment, fix the rest of them + 

Menu is not scrollable when there is not enough space
https://github.com/mate-academy/layout_miami/pull/2036/files

PR #2039, opened 2023-02-04
header added

milenadiamonds1:


OleksiiBlezniuk (changes requested):
Wrong font styles and background color
Screenshot 2023-02-08 at 14 04 58

Wrong outline on inputs, please double check with input states design
Screenshot 2023-02-08 at 14 08 13

Page shouldn't reload on submit, instead - inputs should be cleared to simulate form successful submit

Other layout is nice, also awesome effects on scroll 🔥

src/styles/blocks/container.scss
```diff
@@ -0,0 +1,16 @@
+.container {
+    max-width: 1020px;
+    box-sizing: content-box;
+    overflow-x: hidden;
+
+    margin: 0 auto;
```
kuznetsova-anastasiia, line 6: margins for bem blocks?

kuznetsova-anastasiia (approved):
Amazing job! I like it :)

- image of heading loads slowly
image

- as i see the sides of icon are  rounded in the mockup
your:
image

the design:
image

OleksiiBlezniuk (approved):
Would be nice to make animation on scroll a bit faster

Layout is good, well done!

PR #2040, opened 2023-02-04
adaptive header was added

rashyd-hasratov:
[DEMO LINK](https://rashyd-hasratov.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Phone number should be visible on hover in header
Links in navigation should redirect to existing blocks on page

Wrong font weight
Screenshot 2023-02-08 at 14 16 56

Other layout is good, continue please

KuharchukDmytro:
on mobile distance between sections must be 92px, distance between header and first title must be 60px, distance between footer and contact-us must be 120px, but you have 80px
image

on tablet, phone and menu icons size diffenrent to mockup size
and also on my opinion its a bad way to center them by bottom padding, because there is a big clickalable zone under them, which i think not good

on desktop wrong margins between sections, double check the mockup, please

on my opinion, in forms better to disable autocomplete

KuharchukDmytro (approved):
good job!

PR #2041, opened 2023-02-04
miami_MET-header: first commit

olena-hudimenko:
[DEMO LINK](https://olena-hudimenko.github.io/layout_miami/)


olena-hudimenko:
I could't solve the problem with the implementation of links to 2 phone numbers. It looks strange now, I will go to the Q&A with this question and solve it by the next PR, I hope))

DarkMistyRoom (changes requested):
If you talk about these links, now they work well
image
Keep going and make the next request after finishing the whole job

olena-hudimenko:
I mean, the link only leads to one number out of two

PR #2044, opened 2023-02-05
The MET Landing Page

NikolaienkoN:
https://NikolaienkoN.github.io/layout_miami/


DarkMistyRoom (changes requested):
- the logo shouldn't link to 404
image

- check indent between these icons
image

- the logo in the menu also should be a link
image

- these indents should be equal
image
image

- make a smooth scroll

- fix this overlapping. the left image scale ok, but the right one jumps
https://www.loom.com/share/08bb37a167cc445bbf2c607f73a14890

- same here
image

- add hover effect to these images too
image

- phones and addresses should be links
image

- fix font
image

- all form fields should be required
- there shouldn't be any jumping on form submitting, just reset the form
- check this indent on the mobile
image

Volodymyr-Melnyk-Fron10der (approved):
Good job✅

PR #2045, opened 2023-02-05
made lending bike new versin

IvanVaverchak:
[DEMO LINK](https://IvanVaverchak.github.io/layout_miami/)


IvanFesenko (changes requested):
The first section should have 100% height for any screen
2023-02-06_15h20_38
Click on phone should start a call on device without opening new tab

Location-related addresses / links should open google maps in a new tab target="_blank"

Also add hover effects to this section
2023-02-06_15h24_12

Viktor-Kost (approved):
Well done!

Haven't found what should be happen if I click at **"BOOK A TEST RIDE"**. So i suggest o be scrolled to contact section

PR #2053, opened 2023-02-06
Bike Landing

yaroslav-vasilyev:
[DEMO LINK](https://yaroslav-vasilyev.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Header should have same height as viewport

Scroll should be disabled when menu is opened

Links shouldn't redirect to page with error

Other is good

yaroslav-vasilyev:
> Header should have same height as viewport
> 
> Scroll should be disabled when menu is opened
> 
> Links shouldn't redirect to page with error
> 
> Other is good

Could you explain problem with links?

Yekatery:
image
fix grid
image
Change photo

Yekatery:
Rest - very good!
Telephone animation - top!

Yekatery (approved):
Good job!

Andr1yK (approved):
add transitions
image

PR #2058, opened 2023-02-07
lint fix

Lilu580:
[DEMO LINK](https://Lilu580.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Secondary headings should be bigger on desktop
Links in menu shouldn't be blue
Scroll should be disabled when menu is opened
Screenshot 2023-02-08 at 18 16 25
Menu icon is cropped on hover

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,85 @@
...
+
+    &-image {
+      height: 16px;
+      width: 130px;
+
+      @include onTablet() {
+        height: 24px;
+        width: 190px;
+      }
+    }
```
kuznetsova-anastasiia, line 39: if `header__logo-image` is an element of header then move styles to `.header` using `&__logo-image`, it would be more understandable when you're using bem
if this image is a modificator than it should be `&--image`

kuznetsova-anastasiia (changes requested):
Nice job, you so much work. Let's fix some problems before we go forward)

- favicon looks so bad
image

- phone icon in header should redirect on phone call

- logos redirect to 404 page
image

- make hover effect on this link in menu too
image

- why your names in menu are different from the mockup?
YOUR: 
image

THE DESIGN:
image


- the heading "Browse Bose products by category" is wider than the mockup
YOUR:
image

THE DESIGN:
image


-gap between the last element in gallery and the button is bigger than it should be (your is 64px and the mockup's is 40px)
image


-color of button on hover should be white or transparent
YOUR:
image

THE DESIGN:
image


-same thing on hover
image

- the form button doesn't send the information
image

- as i remember the phone hover don`t have to work on tablets
image

- the heading
YOUR:
image

THE DESIGN:
image

- header image loads a few seconds on mobile

- phone hover again
image

- same thing 
image

kuznetsova-anastasiia (approved):
GJ! fix the issues i described below asap, before mentors will see them)

- this links don't have hover effects
image

- on tablet paddings have to be 72px
image

and on desktops they have to be 120px

OleksiiBlezniuk (changes requested):
Well done, but please fix @kuznetsova-anastasiia comments

OleksiiBlezniuk (approved):
Well done, there are few issues that are not critical, but better to fix them

Scale on hover looks too big, I think better to make it a bit smaller, but it's not critical

`View all` button redirects to top of the page and it's not logical, better to redirect to other page block or third party resource

PR #2063, opened 2023-02-07
header mobile was created

vlad-fil:
[DEMO LINK](https://vlad-fil.github.io/layout_miami/)


IvanFesenko (changes requested):
Nice start but it's only mobile view. + Click on logo and menu leads to 404 page.
Add favicon

nazarmatsevych (changes requested):
Good work,
but next time please make re-request when you completely finished your landing page

PR #2068, opened 2023-02-08
Header done

YaroslavDzyadiv:
[DEMO LINK](https://YaroslavDzyadiv.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Don't forget to reset headings default margin and to use semantic tags
Mobile layout is good except too big distance between heading and bottom of the page

OleksiiBlezniuk (changes requested):
Distance between h1 and bottom of the page should be bigger on desktop

Other layout is good, looking forward for final solution

Vlados3611 (commented):
Great job! I found some mistakes on your page, so better to fix them ;)

You need to decrease the block logo height:



Picture from figma:



Also you need descrease the logo height from menu block. Because of this, your panel is located lower than it should be:





Picture from figma:





You need to fix the height of the pictures from 320px and also from 744px. From 320px screen = 280px width, and from 744px screen = 392px width. 

For example from 744px screen width: 



Picture from figma:



And also fix the border, it should be 16px. But actually it's not important :)



Picture from figma:

src/index.html
```diff
@@ -1,13 +1,424 @@
...
+              </h3>
+
+              <p class="bikes__description">
+                The iconic frame brought to a new performance height as a sporty, active ride.
+              </p>
+
+              <h3 class="bikes__price">
+                $ 2 590
+              </h3>
+            </div>
```
Vlados3611, line 180: Actually better to make a width limit for this block 392px starting from 744px of the screen width, like in figma:

Vlados3611 (approved):
Cool, Good job :) Please check my last comment

illiarerun (changes requested):
- [checklist](https://github.com/mate-academy/layout_miami/blob/master/checklist.md) 9th item
- fix your message field border as Vlad suggested
image
- add hover effects to form
image
- add transition to ALL buttons
image

illiarerun (approved):

PR #2072, opened 2023-02-08
MyBike landing

MaksymMaliuk:
[DEMO LINK](https://MaksymMaliuk.github.io/layout_miami/)


src/index.html
```diff
@@ -1,13 +1,197 @@
...
+             class="logo"
+            >
+              <img
+                src="images/logo.svg"
+                alt="MyBike"
+                class="logo__image"
+              >
+            </a>
+            <div class="top-actions">
+              <a href="tell:+1 234 5555-55-55" class="icon--phone icon header__call"></a>
```
vladyslav-mosor, line 28: 'tell' => 'tel'

vladyslav-mosor (changes requested):
1. Disable scroll for the page when your menu is opened. And it`s better also to close the menu automatically when the page is refreshed during the menu opened, but it goes up to you.

2. Your images must be placed in the center on tablets and phones.

3. Your link that is shown when icon call is hovered does not work as link to phone number. Please fix it

sergey-lipikhin (changes requested):
Change color while hovering on phone, email and address. And with transition for color changing (CL p. 4)
image

image


Text of form fields has different styles?
image

Avoid empty values in form fields, according to checklist
image

sergey-lipikhin (approved):

Andr1yK (changes requested):
Why I can't submit form?)
image

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,94 @@
+.header {
+  background-image: url(../images/bitmaps/cover.png);
```
sergii-nosachenko, line 2: Why do you need so a huge image for bg? PNG is for icons or illustrations, but not for photos. Use JPG instead.

src/styles/blocks/footer.scss
```diff
@@ -0,0 +1,14 @@
+.footer {
+  height: 200px;
+  background: url(../images/bitmaps/footer.png) no-repeat center;
```
sergii-nosachenko, line 3: Same issue with file format

sergii-nosachenko (approved):
Good work.
Commented on the chosen image format.

PR #2084, opened 2023-02-08
Mayami landing

Yekatery:
- [DEMO LINK](https://Yekatery.github.io/layout_miami/)

### DIFF
--- a/src/styles/blocks/button.scss
+++ b/src/styles/blocks/button.scss
@@ -24,6 +24,7 @@
     border-radius: 40px;
 
     cursor: pointer;
+    transition: box-shadow 0.5s;
 
     &:hover {
       box-shadow: 1px 4px 16px rgba(232, 232, 240, 0.4);
--- a/src/styles/blocks/contacts.scss
+++ b/src/styles/blocks/contacts.scss
@@ -21,6 +21,7 @@
     font-size: $f-size-text;
     line-height: 140%;
     color: #fff;
+    transition: color 0.5s;
 
     &:hover {
       color: $c-hover-sec;
--- a/src/styles/blocks/form.scss
+++ b/src/styles/blocks/form.scss
@@ -17,10 +17,10 @@
   }
 
   &:hover::placeholder {
+    transition: color 1s;
     color: $c-hover;
   }
 
-  &:hover,
   &:focus {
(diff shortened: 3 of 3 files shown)


OleksiiBlezniuk (changes requested):
Please double check with design, distance between this icons and top of the page should be a bit bigger, icons in the left corner should be smaller
Screenshot 2023-02-10 at 17 47 58

This text should have same width as left image on desktop (same columns in grid)
Screenshot 2023-02-10 at 17 51 53

Links to outer resources should be opened in new tab (link to maps)

`send` button on desktops should be as wide as inputs and it shouldn't redirect user to top of the page. It sohuld submit form (without page refresh, but it should clear inputs to simulate that result is sent)

Add hover effects to images in gallery, to links and to cards

src/index.html
```diff
@@ -1,13 +1,494 @@
...
+    <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Miami</h1>
+  <body class="page__body">
+    <header class="header">
+      <div class="container container__header">
+        <div class="header__content">
+          <div class="top-actions">
+            <a href="/" class="logo">
```
milenadiamonds1, line 27: add # instead of / because i have an error when clicking on the logo(in menu do the same)

src/index.html
```diff
@@ -1,13 +1,494 @@
...
+
+            <div class="top-actions--wrapper">
+              <a
+                href="tel: +1 234 555-55-55"
+                class="
+                  icon
+                  icon--phone
+                  header__call"
+                ></a>
+              <a href="#menu" class="icon icon--menu"></a>
```
milenadiamonds1, line 43: Знімок екрана 2023-02-10 о 20 20 00
some problems with icons, and there are too big as for me, please double check the mockup

milenadiamonds1 (changes requested):
GJ! But fix a few problems, and also your landing will look som much better with animations)

milenadiamonds1 (approved):
Well done!

illiarerun (changes requested):
add transition-duration to this button:
image
add transition-duration to form:
image
and transition-duration here:
image

illiarerun (approved):

PR #2089, opened 2023-02-08
MyBiKE

VadKol:
- [DEMO LINK](https://VadKol.github.io/layout_mybike/)

- [MYBIKE landing figma](https://www.figma.com/file/NZQAIydtHo5QkINyGLHNcq/BIKE-New-Version?node-id=0%3A1)

- [CHECKLIST](https://github.com/mate-academy/layout_miami/blob/master/checklist.md)


Stepan-TS (changes requested):
1. pay attention to first block "move free".  It must be at the begin of the page; also your paddings (top and bottom) correct for only mobile version) 
move free
move free grid example

2. indents on the sides should be 120px;
margins
true margins

3. your text in gallery must be in the first 8 grid columns;
text
true text


4. your contacts not correct. inputs must be in the first 5 grid columns, also turn off default outline.
contacts
true contacts

yuliia-vereshko (changes requested):
1. Fix your menu on small screens 
menu on small screen
2. Your sides are not as on mucket. check it.
margin on tablet screen
3. Font size is bigger than on your site on the desktop screen.  
font-size
4.  Change text color on hover for phone, email and address.
font color in hover
5. Align text on Desktop screen
text align
6. Gap between cards must be 24px on Desktop .
gap between slide

Ambulence (approved):
Not bat! Just pay attention to the form. Original markup doesn't have box-shadow on hover and outline on target. Another is good

Andr1yK (approved):
IMHO, this image should be full screen
image

PR #2090, opened 2023-02-08
implemented adaptive header and nav-menu

Ambulence:
-[DEMO LINK](https://Ambulence.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Layout is good, looking forward to see final solution
Don't forget to change title, set favicon and add hover effects to images in gallery, to cards and to links

valeriiayarzhemska:
image
Text isnt centerned

valeriiayarzhemska:
image
Text isnt centered

valeriiayarzhemska:
image
On tablet u have different position of bg

valeriiayarzhemska:
image
U have outline on hover and target

valeriiayarzhemska:
image
Outline on hover

src/styles/blocks/nav-menu.scss
```diff
@@ -0,0 +1,73 @@
...
+
+    font-weight: 600;
+    font-size: 22px;
+    line-height: 22px;
+    color: $menu-item-color;
+    text-decoration: none;
+    letter-spacing: 2px;
+    text-transform: uppercase;
+
+    transition: border-color $transition linear;
```
valeriiayarzhemska, line 41: as i remember, its better to add transition also on hover too (not only in this part of your code)

src/styles/blocks/about.scss
```diff
@@ -0,0 +1,17 @@
+.about {
+  &__title {
+    text-align: start;
+  }
+
+  &__info {
+    margin: 32px 0 0;
```
valeriiayarzhemska, line 7: better to add just margin-top (it applies to all of your project)

valeriiayarzhemska (changes requested):

Ambulence:
> 



> image Text isnt centered

image

src/styles/blocks/nav-menu.scss
Ambulence: I think not, if I add a transition on hover too it will work to one side

src/styles/blocks/bike.scss
```diff
@@ -0,0 +1,53 @@
+.bike {
+  &__photo-container {
+    width: 100%;
```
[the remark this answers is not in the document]
Ambulence: I don't think so, for example, I have here width 100%, and in another part of my code I have translateX 100%,  and in another, I have right 100%. I make one variable with the name "hundred" or "hundred percent", ok. Then I need to change the width to 50%, I change the variable to 50% BUT I want to leave translateX 100% and right position 100%. So I need to make one more variable again or change my code in a few places. Seems not cool

Ambulence:
> image Text isnt centered


it seems))

valeriiayarzhemska (approved):

illiarerun (approved):
phone looks weird on desktop:
image

PR #2093, opened 2023-02-08
created header and mobile menu

valeriiayarzhemska:
[DEMO LINK](https://valeriiayarzhemska.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
I don't think that it's a good idea to use percents for that purpose, it's enough to set margin from mock and make it stable and not depended on screen height
Screenshot 2023-02-09 at 17 50 46

Distance between these icons should be a bit bigger on desktop, check on design
Screenshot 2023-02-09 at 17 53 51

Better to limit main content width to be sure that layout will be good on even bigger screens (it's limited on design btw)

Heading font size should be bigger, wrong distance between text in left block on desktop
Screenshot 2023-02-09 at 17 56 30

Also content of this block shouldn't be displayed in column on tablets

Mobile layout is good except wrong margins in text block (same as on desktops)

OleksiiBlezniuk (changes requested):
Page shouldn't have horizontal scroll

Don't forget to change title, add favicon and hover effects to images, cards and links

kateryna-babina:
image
This phone button doesn't have hover effect

kateryna-babina:
image
These menu icon and close icon are not on the same level

kateryna-babina:
image
you should change logo behavior on click, try to add href= "#"

kateryna-babina:
image
This subtitle should be in two lines

kateryna-babina:
image
This button doesn't work :(

kateryna-babina:
image
This title also should be in two lines on tablets

kateryna-babina:
Check the mobile version, because I found a horizontal scroll there

kateryna-babina (changes requested):
Good job, but fix some issues! :)

kateryna-babina (approved):

Andr1yK (approved):
🫠

https://user-images.githubusercontent.com/86183007/219501575-f3a67113-d0e4-49b1-9c6e-da5723840fa9.mov

Too big scale
image

PR #2095, opened 2023-02-09
Layout Bike Landing

ArthurKrepko:
[DEMO LINK](https://ArthurKrepko.github.io/layout_miami/)


OleksiiBlezniuk (changes requested):
Header layout is good, looking forward to see full solution

Andr1yK (changes requested):
For icons better to use svg files
image

Let's add some animations on hover
image
image

IMHO, those fields should be required
image

Points 13, 14, 15
image

And.. check by checklist, pls)

PR #2102, opened 2023-02-10
My Bike lp was created

Tarasova09:
-[DEMO LINK](https://Tarasova09.github.io/layout_miami/)
There is no animation on scroll yet.


Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Great work, but there are some issues

1. Implement checkpoint
image
image
2. Form should be cleared after submitting
image
3. logo in the menu leads to 404 error
image

nazarmatsevych (approved):
Great work!

IvanFesenko (changes requested):
Cool work. One small remark - you should disable scroll in case when menu is shown
Just use this code:
Styles:
```
.page__body--with-menu {
  overflow: hidden;
}
```
JS:
```
window.addEventListener('hashchange', () => {
  if (window.location.hash === '#menu') {
    document.body.classList.add('page__body--with-menu');
  } else {
    document.body.classList.remove('page__body--with-menu');
  }
});
```

mykhalenych (changes requested):
the layout breaks on hover
https://user-images.githubusercontent.com/53228259/218505827-067baa84-308b-43f7-b23a-12b315fde2d2.mov

OlEzhA0 (changes requested):
Good job! Please provide few improvements
1. click on logo refreshing the page
2. Something wrong with effects
https://user-images.githubusercontent.com/60295799/218694638-f3f4c33d-166a-4087-9089-2bf8af346731.mov

Viktor-Kost (approved):
Cool!)

P.S. I would recommend to change a hover effect on logo. According to UI you'd better not use rotating on logo. But, anyway it's up to you)

Tarasova09:
> Cool!)
> 
> P.S. I would recommend to change a hover effect on logo. According to UI you'd better not use rotating on logo. But, anyway it's up to you)

Yes, you are right, but what should I use instead? I tried scale and shadows, but they were not good  also. It was other kind of rotation, but there was some strange problem and I still don't know how to fix it.

VitaliyBondarenko1982 (approved):

PR #2123, opened 2023-02-16
Header, footer, menu

khomenkom:
[DEMO LINK](https://khomenkom.github.io/layout_miami/)


tetiana-ivanchenko (changes requested):
re-request the review when the landing will be ready

qoa11a (changes requested):
demo link doesn't work. Please fix it. If you need help ask in chat in Slack

khomenkom:
Demo link fixed

qoa11a (changes requested):
good job! 👍 
few comments:

1. handle form submission correctly. It shouldn't reload the page, shouldn't redirect to 404 and inputs should be verified, and cleared after submission
2. wave icon on mobile is cropped
image
3. the image shouldn't be this big and cropped as well on mobile
image
4. here text should be 16px, not 24px
image

please, fix all these mistakes and also look for more differences, there are more. Compare your solution to the design closely and pay attention

OleziO (changes requested):
Demo:
image
Design: 
image

Demo: 
image
Design:  
image

Demo: 
image
Design: 
image

PR #2127, opened 2023-02-18
Met landing

iamgranovskaya:
[DEMO](https://iamgranovskaya.github.io/layout_miami/)


OleziO (changes requested):
Landing is not completed.

PR #2129, opened 2023-02-18
header and menu slider added

daryna-hnidash:
in progress

[DEMO LINK](https://daryna-hnidash.github.io/layout_miami/)


etojeDenys (approved):
looks good,

nit: `Same when you click on logo. There shouldn't be any error.`

vlad-demchuk (changes requested):
checklist
image

YegorKorenkov:
Please fix start position in textarea
image

etojeDenys (changes requested):
LGTM

1. Form shouldn't be submitted if some of the fields are not filled
2. let`s fix that 
image
3. form values should reset on submit

YegorKorenkov:
Please fix 404 error when you click on home
image

YegorKorenkov:
Hover don't work
image

daryna-hnidash:
> Hover don't work image

It doesn't work on purpose cause it`s tablet version under 1024px and on tablets we don`t have hover. Am I right?

YegorKorenkov:
> > Hover don't work image
> 
> It doesn't work on purpose cause it`s tablet version under 1024px and on tablets we don`t have hover. Am I right?

Yes u are right. Usually, the hover is applied in general for all sizes, if for example the user will resize the browser window, but these are already my nit-picking. Nice work!

PR #2157, opened 2023-03-01
header was created + desktop + tablet

olyakachmarchyk:
https://olyakachmarchyk.github.io/layout_miami/


vlad-demchuk (changes requested):
logo leads to 404 now and use svg for it|

re-request when you implement the next block

IvanFesenko (changes requested):
2023-03-03_14h08_59
Your page not working

VitaliyBondarenko1982 (changes requested):
Good start.
We are waiting next blocks.

OlEzhA0 (changes requested):
gj! waiting for next blocks

vlad-demchuk (changes requested):
Looks good

Keep doing

etojeDenys (changes requested):
1. fix this second line 
 image
2.  open address links in a new tab 
image
3. rewrite button styles follow 
image
4. try to make this button at the same height as in the menu
image
image

etojeDenys (changes requested):
1. I can see the whole page only after clicking on the logo
2. Form shouldn't be submitted if some of the fields are not filled
3. Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)

OlEzhA0 (changes requested):
check this bug
Screenshot 2023-03-16 at 11 22 16 AM

etojeDenys (changes requested):
i still can see this overlapping
image

olyakachmarchyk:
> i still can see this overlapping image

sorry for the inconvenience, will fix

etojeDenys (approved):
awesome 🤩

PR #2163, opened 2023-03-03
MyBike created

GeorgeSavostikov:
[DEMO LINK](https://GeorgeSavostikov.github.io/layout_miami/)


DarkMistyRoom (changes requested):
Good job. Let's improve a bit
- fix the textarea font
image

-  page shouldn't reload on the form submitting, just reset the form
- download the image directly from the mobile screen on the figma to make it looks like on the mockup
image
image

- add indent here(mobile)
image

- and it is good practice to download images in better resolution. to make them look good on scale
image

Volodymyr-Melnyk-Fron10der (changes requested):
keep it up💪🏼 I see only one issue is left actually

1. check image scaling on the mobile
image

Volodymyr-Melnyk-Fron10der (approved):
ThumbsUpWellDoneGIF

PR #2168, opened 2023-03-03
landing header

theFallcon:
[DEMO LINK](https:/theFallcon.github.io/layout_miami/)


Anastasiia-Svintsova (changes requested):
- Compare the size of the icons with the design, please 
image

- Please, implement the whole page, before requesting a review again

agniya-i (changes requested):
Implement the whole page, before requesting a review again 

Also, pay attention that now all links for the menu follow to 404 page, it should be anchor links

OlEzhA0 (changes requested):
1. page shouldn’t refresh after click on Send in form
2. check this bug
Screenshot 2023-03-08 at 10 13 08 AM
3. seems like first screen should be 100vh
Screenshot 2023-03-08 at 10 14 14 AM

polosanya (changes requested):
I still can scroll while being in burger menu
image
image

OlEzhA0 (changes requested):
check this bug
Screenshot 2023-03-10 at 12 29 56 PM

sTorba24 (changes requested):
Remove x-scroll on mobile 
image

content shouldn't be visible when menu opened 
image

agniya-i (changes requested):
Please fix one of the previous comments, It would be better if the header had height 100vh 
Screenshot 2023-03-15 at 21 22 51

loralevitska (changes requested):
Almost done) Let's improve your menu
- Don't teleport user by click on logo in menu. Just smoothly close menu (same with open menu)
- Remove this. Try to download picture from figma without this block
image

PR #2202, opened 2023-03-06
start develop

Rage1502:
[DEMO LINK](https://Rage1502.github.io/layout_miami/)


danylo-dobrovolskyi (changes requested):
I think you should check the checklist for this task one more time, especially paragraphs number 4(you need to change text color on hover), 6(there is '404 error' when I click on 'bose' logo, 7(it opens but the location is wrong), 12(I'm not sure that the animation speed is the same everywhere, but if so, then skip this paragraph). Sasha said that we don't need to look through the code so this is everything that I could find on the first look(but there are no big mistakes, you need to do it with checklist rules)

Rage1502:
> I think you should check the checklist for this task one more time, especially paragraphs number 4(you need to change text color on hover), 6(there is '404 error' when I click on 'bose' logo, 7(it opens but the location is wrong), 12(I'm not sure that the animation speed is the same everywhere, but if so, then skip this paragraph). Sasha said that we don't need to look through the code so this is everything that I could find on the first look(but there are no big mistakes, you need to do it with checklist rules)

SerhiiKirik (changes requested):
Great 🚀 
Let's fix some problems & make some improvements 🧑‍💻 
https://www.loom.com/share/986268a34c69467dafe2d9b0e28aeafc

Rage1502:
> Great 🚀 Let's fix some problems & make some improvements 🧑‍💻 [loom.com/share/986268a34c69467dafe2d9b0e28aeafc](https://www.loom.com/share/986268a34c69467dafe2d9b0e28aeafc)

Thanks for your comments, I've already improved everything.

SerhiiKirik (approved):
LGTM 🔥 

Also, you can make some improvements 🧑‍💻 

Add target="_blank" here & add styles accordenly to the design
image
image

Remove all hovers on phone & tablet
image

PR #2222, opened 2023-03-07
first commit - header

glots-a:
https://glots-a.github.io/layout_miami/


VenomMB (changes requested):
Good Job, but you have mistakes, please watch this video.
https://www.loom.com/share/54d3d4aee30a4844830b5352192d2f16

SerhiiKirik (changes requested):
Great 🚀 
Let's make some improvements 🧑‍💻

I can send message without a massage, fix it
image

put max-width to the container this section
image

image


don't use all, it's too hard for the browser
Just list params, for example, tansition: color 0.3s
 image

Different font size
image

image

It would be better if you add a new class for it & put overflow: hidden
image

glots-a:
hello, as to the last phrase - I dont anderstand where and why I should to add class with 'overflow: hidden'

PR #2235, opened 2023-03-08
Develop

LeoSpartaDeV:
https://leospartadev.github.io/layout_miami


SerhiiKirik (changes requested):
Well done 🚀 
Keep it up

Volodymyr-Melnyk-Fron10der (changes requested):
good job👍 

1. scale all photos on hover
image
2. add favicon
image
3. page shouldn't be scrolled if menu is open
image

LeoSpartaDeV:
> good job👍 
> 
> 
> 
> 1. scale all photos on hover
> 
> image
> 
> 2. add favicon
> 
> image
> 
> 3. page shouldn't be scrolled if menu is open
> 
> image
> 
> 

Thanks😀 done

DarkMistyRoom (changes requested):
Keep going! Here are a bit more suggestions:
- all these links should lead to the appropriate section of the page

- add hover effect to these images too
image

- make the smooth scroll on your page

- fix the font
image

- make this button the same height as previous one. and fix the style. there is example of hovere state on the mockup. also cursor should become a pointer over it
image

- add hover effect to these links. and make address link lead to google maps
image

- there should be some minimal side paddings
image

PR #2239, opened 2023-03-09
header mobile version-1

Nikolaymelni:
Hello! This is a header for mobile version 320px
I have two question, is this right, that there are 2 different pictures for header (for mobile and desctop)?
And I made one more <div> for header because I need to put the icons differently, not like in the video.
could you advise how to do it better? because here I had to specify the exact distance between the icons

https://Nikolaymelni.github.io/layout_miami/


anstsot (changes requested):
Good start! Re-request the review once you're done whole landing.

Nikolaymelni:
Hello! I finished this task, looking forward to review

Nikolaymelni:
I found one mistake. somehow margins between text and images in "More to explore online" don`t work. I tried to fix it but it doesn`t work. Could you please explain me why this doesn`t work?

code from block "explore"
&__description {
    font-family: "Roboto", sans-serif;
    font-size: 16px;
    line-height: 27px;
    color: $black-color;
    text-align: center;
    margin: 0 0 48px;

    &:last-child {
      margin: 0;
    }

  }

anstsot (changes requested):
Good job, but there are few fixes to improve your code.
1. On logo click I'm ending on 404 page. it should stay on the same landing page.
2. After submitting the form the whole page was reloaded. it should stay on the same place and the form should be cleared.
image

3. add the links in contacts section
image

4. please add margin on smaller device
image

Nikolaymelni:
Thank you for review. Fixed everything. Googled what to do with form and as I found, I can change behavior of form only with JS code. Right?

polosanya (changes requested):
Great! Almost done!
1. It would be great to add a hover effect for these images too
image
2. Same with these links. Add some hover effects for them. 
Also, make sure that Google Maps links are opened in a new tab
image

polosanya (approved):
Well done!

Anastasiia-Svintsova (commented):
- The title position seems too high 
image

- The title width doesn't match the design on Tablet 
image

- The gap between sections doesn't match as well
image

PR #2241, opened 2023-03-10
MyBike - first blocks

AnnaNazarova23:
DEMO LINK - https://AnnaNazarova23.github.io/layout_miami/


etojeDenys (changes requested):
good job

1. Location-related addresses / links should open google maps in a new tab target="_blank"
2. Form shouldn't be submitted if some of the fields are not filled
3.  fix this overlapping
image

PR #2243, opened 2023-03-10
header-without-menu

AnastasiiaVorobets:


VitaliyBondarenko1982 (changes requested):
Good start.
Click on logo leads on 404 page.
check background-position for header backgroung image.
on some width too hard to see numbers on phone
Screenshot from 2023-03-10 20-46-15

We are waiting other sections.
Please re-request when all sections will be implemented.

sTorba24 (changes requested):
Great job :) lets make it perfect :)
It should be aligned on the same level
image

it's redirecting me on 404 page(check all links to avoid it)
image

fix margin here (for all devices)
image

submit button also redirecting to 404 page(fix it and would be good to clear inputs after click on submit using js )

vlad-demchuk (changes requested):
check this checklist points (all fields should be required and form submission should just clear fields)
you will need some js code, if you have any questions, just feel free to ask in `fe_chat`
image

slavanosatskyi (changes requested):
Nice work, though some polishing is needed.
1. Logo leads to 404 page
Screenshot 2023-03-13 at 17 28 15
2. Please, add hover effects for emails, address, etc. (color change for example)
3. Map should be opened in the new tab
Screenshot 2023-03-13 at 17 28 46
4. Please, add form validation, so it is no possible to submit empty form or form with non-valid email
Screenshot 2023-03-13 at 17 29 29

vlad-demchuk (changes requested):
1. now email is a link to `tel:`
image
2. button should have `type="submit"`, in this way your validation doesn't make any sense
image

if you have any questions, just feel free to ask in `fe_chat`

YegorKorenkov (changes requested):
1. Page shouldn't be reloaded on form submit
2. A little bit centred this block
image

PR #2253, opened 2023-03-12
solution

Sergej161995:
[DEMO LINK](https://Sergej161995.github.io/layout_miami/)


tetiana-ivanchenko (changes requested):
[checklist](https://github.com/mate-academy/layout_miami/blob/master/checklist.md)

1. fix scroll
image
3. all hover effects should have transition
4. fix checklist point 1
5. fix checklist point 4
6. fix checklist point 8
7. fix checklist point 10
8. fix checklist point 14
9. fix checklist point 15
10. 
image

AlisaVlasova (changes requested):
1. use svg 
image
2. add smoth transition to all hovers
image
3. fix textarea 
image
4. reset form fields on submit and remove autocomplete styles
image

etojeDenys (changes requested):
1. fix this overlapping
image
2. 
image
image
3. Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)

TarasHoliuk (changes requested):
GJ :+1: 
Only a few things should be fixed

mykhalenych (changes requested):
missed requirement

1. Form shouldn't be submitted if some of the fields are not filled

also check design
it is design
Screenshot 2023-03-17 at 17 06 20
it is your
Screenshot 2023-03-17 at 17 07 19
section with "Why buy direct" different too

PR #2255, opened 2023-03-13
make header

NadieinOleh:


sTorba24 (changes requested):
Good start, keep it up:) and request review when you finish

qoa11a (changes requested):
image
image
- Compare your solution to the design. Pay attention to the background image and burger menu icon.
- Disable scrolling when the menu is open.
- Click on phone shouldn't redirect to 404.
image
image
- Again, compare your solution to the design. Pay attention to images. Check it everywhere, not just in these two sections. You have problems with every image.
- Align inputs and a button in the form.
- The form submission should clear inputs and it shouldn't scroll the user to the top of the page.
- All links should work and be opened in a new tab. Some of them direct to a 404 page.
image
image
- Again, different layouts in your solution and the design.
image
image
- Please, check your solution to the design. Pay attention, and watch closely. There are a lot of differences.

polosanya (changes requested):
Looks great! Here are some comments to fix:
1. Burger menu background color doesn't match the design
image
image
2. Burger menu icon doesn't match the design. Also, it would be great for all interactive elements to have cursor pointer on hover
image
image
3. Let's increase the images on hover
image
4. Same with these images
image
5. You can download it without this button as follow:
image
image
6. Add some hover effects for these links
image
7. Fix it
image

OleziO (changes requested):
Demo: 
image
Design: 
image

Demo: 
image
Design: 
image

OleziO (approved):

PR #2256, opened 2023-03-13
Develop

danielpopek94:
[DEMO LINK](https://danielpopek94.github.io/layout_miami/)


michalcywka (changes requested):
- The form should send the data somewhere (doesn't matter where)
- The button should have width of the whole content on small screen (see screenshot)
Screenshot 2023-03-27 at 16 12 58
- On large resolution screens we want the content to be centered, not streched. Use `max-width` of 1260 (see screenshot)
Screenshot 2023-03-27 at 16 15 59

michalcywka (changes requested):
On large resolutions, the navbar should be as wide as the rest of the content (right now it is on the edges)
Screenshot 2023-03-30 at 13 26 29

d-wasilewski (approved):

PR #2273, opened 2023-03-18
layout_miami add task solution

denys2:
[layout_miami solution URL:](https://denys2.github.io/layout_miami/)


polosanya (changes requested):
1. Background color doesn't match the design. Also, check the distance between logo and first link. By the way, looks like font is also different
image
image
2. Let's download the image without that button in the corner
image
3. This button does nothing on click. Navigate somewhere or open some external source 
image
4. Make sure that this link is opened in a new tab
image
5. Let's add some validation to the form (required, email format)
image
6. It doesn't look correct. Add some paddings 
image

denys2:
I corrected the comments

AgniyaIvanishyna-GBI (changes requested):
Please check font weight and font size for all elements, it looks different from the Figma design

For example: 
Screenshot 2023-03-30 at 17 22 13
Screenshot 2023-03-30 at 17 22 40

2. Form submit should not open a new page, better reset fields (checklist p15)

polosanya (changes requested):
Almost perfect! One problem is still not fixed
1. Background color of burger menu still doesn't match the design
image
image

polosanya (approved):
Well done! 🔥

PR #2288, opened 2023-03-22
The Met solution

bjeziorek:
-  [DEMO LINK](https://bjeziorek.github.io/layout_miami/)


michalcywka (changes requested):
Very nice, just two things:
- Links in the menu should have some kind of hover effect (for example changing colour or underline)
- The "View all exhibitions" button should take us either to the top of the page or to some external web page.

michalcywka (approved):

PR #2289, opened 2023-03-22
Develop

Dorosh90:
-[DEMO LINK](https://Dorosh90.github.io/layout_miami/)


DarkMistyRoom (changes requested):
Great job. Fix a few moments:
- fix the logo link. now it leads to the 404
image

- the close button in the menu and menu button in the header should be on the same position https://www.loom.com/share/474e80cc7de8497294de049e151da7b8
- make the message required too. and don't reload the page on form submitting
image

- use the image directly from the mobile to make it looks good
image

sTorba24 (changes requested):
Good job! Some things to be improved 

1. fix background image 
image
you can try to play with background-size: contain(it'll broke view for start, but generally image will be positioned better :)

2.image is overlapping text on hover
image

sTorba24 (changes requested):
background behaviour can be better 
image

it'll be checked by recruiter and sometime this background isn't well positioned

sTorba24 (changes requested):
try to set 
background-position: 50% for >744
before background-position: 85% 50%

mykhalenych (approved):

marina-tilniak (approved):

PR #2290, opened 2023-03-22
Develop

Dka888:
[DEMO LINK](https://Dka888.github.io/layout_miami/)


michalcywka (changes requested):
- Form should send data somewhere (also sometimes when i click on Send button it opens the side menu)
- On very wide resolutions, the nav buttons should be centered just as the rest of the content
- Give the footer image `background-size: cover`
- I would expect the telephone numbers in the side menu to also make a call action

michalcywka (approved):

PR #2293, opened 2023-03-22
The MET

Ozwiena:
[DEMO LINK](https://ozwiena.github.io/layout_miami/)


michalcywka (changes requested):
- Use higher resolution images
- In Contact Us section, all the information should have some action: e-mail tries to send an e-mail, telephone number tries to call, address takes us to Google Maps with this location
- On high resolution screens, the content should have be centered and have max-width
Screenshot 2023-03-28 at 10 50 28
- Telephone icon next to menu should try to call. Same with telephone numbers in menu and Call to order button
- Use `background-size: cover` for footer (now it looks bad on some mobiles)
- Scrolling menu on mobile when in landscape mode does not work
Screenshot 2023-03-28 at 10 57 48

michalcywka (changes requested):
There is one small problem, when on mobile and in landscape mode, the menu is not being scrolled properly. (see that scrollbar on the right side in the video is moving but the content is not)
https://user-images.githubusercontent.com/23585899/228240127-8dc887dc-0bf1-4703-82a7-719224187a43.mov

michalcywka (approved):

PR #2299, opened 2023-03-22
Develop

sylcym:
[DEMO LINK](https://sylcym.github.io/layout_miami/)


michalcywka (changes requested):
- Add a cursor change when hovering over images e.g. `cursor: zoom-in;`

michalcywka (approved):

PR #2302, opened 2023-03-22
MET 

PufAnimalsDev:
[DEMO LINK](https://pufanimalsdev.github.io/layout_miami/)

### DIFF
--- a/src/styles/components/header.scss
+++ b/src/styles/components/header.scss
@@ -24,14 +24,11 @@
     letter-spacing: -1.5px;
 
     @media (min-width: 600px) {
-      padding-top: 141px;
-
+      padding-top: 200px;
       font-size: 64px;
     }
 
     @media (min-width: 1024px) {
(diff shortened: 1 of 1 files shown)


kasiakubisiak (changes requested):
I just have one thing - favicon isnt visible. Except that - good work!

kasiakubisiak (changes requested):
add object-fit: cover; to images inside More to Explore Online section

kasiakubisiak:
except that one change it looks great! 🥳

kasiakubisiak (changes requested):
I'm not sure if it was like this before but I have one more improvement to make:
- put "Welcome to The MET" text in hero section in the middle of the image or lower it a bit because now it covers artwork's face on desktop view 😄 
image


and that would be all! let's fix that one thing and it will be perfect, good job!

kasiakubisiak (approved):
good job! 🔥

PR #2304, opened 2023-03-22
Develop

kq333:
[DEMO LINK](https://kq333.github.io/layout_miami/)


michalcywka (changes requested):
-Header should take height of 100vh
-On small screens, text is truncated in contact form. It should go below the form
Screenshot 2023-03-27 at 17 12 31
- On large resolutions the header image should also take whole height of the screen
Screenshot 2023-03-27 at 17 22 46

michalcywka (changes requested):
- The header and footer images should always take 100% width
Screenshot 2023-03-28 at 10 24 47
Screenshot 2023-03-28 at 10 24 51

michalcywka (changes requested):
- Call links don't need to open a new tab when clicking on them
- The navbar should have some kind of max-width, just like the content.

kq333:
> 
According to check list no. 9 ('Location-related addresses / links should open google maps in a new tab target="_blank" ')

michalcywka (changes requested):
The navbar on large resolutions should be as wide as the rest of the content (now it is at the edges of the page)
Screenshot 2023-03-30 at 13 24 50

michalcywka (approved):

PR #2305, opened 2023-03-23
Develop

OleksChernikov:


qoa11a (changes requested):
- Add validation for form inputs and clear inputs after the form submission.
- Open all links in a new tab.
- Download MyBike logo as an SVG.
- Align images on the tablet.
image
- Compare your solution to the design on mobile (320px). Fix it.
image
image
- Why do you have another background image for the hero section?
image

PR #2307, opened 2023-03-23
add push Bike

6houston6:
[DEMO LINK](https://6houston6.github.io/layout_miami/)


polosanya (changes requested):
Looks really great! Let's improve your page a bit:
1. Background color of burger menu doesn't match the design
image
image
2. Download the image without this button
image
image
3. Add some hover effects for links
image
4. Make sure that this link is opened in a new tab
image
5. Form shouldn't submit with empty fields. Add some validation.
Also, don't pass the data through URL
image
image
6. Main screen on tablets doesn't match the design
image
image

anstsot (changes requested):
Great job!
Just one point left. Please check the form. it shouldn't be reloaded after submit
image

PR #2308, opened 2023-03-24
done first try, ready for million change requests

VladyslavKucher:
https://VladyslavKucher.github.io/layout_miami/

I know you will have a million requests. However, if it is about grid, please give me more details how to fix it. :)


polosanya (changes requested):
Nice start! Let's improve your page a bit)
1. Add favicon for the page
image
2. There should be no error on logo click
image
3. Burger menu icon doesn't match the design
image
image
4. Add increasing hover effects to all images
image
image
5. Fix indentation between inputs. Also, try not just navigate on submit, but at least add some validation to inputs (required, email format)
image
6. Add hover effects to these links. Make sure that Google Maps link is opened in a new tab
image
7. Doesn't match the design on mobiles
image
image

Volodymyr-Melnyk-Fron10der (changes requested):
1. logo in menu leads to the 404 error
image

2. that's too far, do not overlap other cards 
image

3. button should validate inputs and reset the form if all correct, not scroll to the top
image

4. contacts are placed in the wrong place on mobile view
image

5. only image should be increased 
image

witflash (changes requested):
Almost is done 👍🏻
Could you do the last small fixes?

1. Remove page scroll when mobile menu is opened. Scroll should be just for menu not for the whole page.
2. Could you check if your fav icon has a cross browser solution? At least I don't see it at iPad for instance.
3. Check the image hovers. I see on some breakpoints it moves a little near text blocks. I think, if you will use `transform: scale` it will be fixed.

VladyslavKucher:
> Almost is done 👍🏻 Could you do the last small fixes?
> 
> 1. Remove page scroll when mobile menu is opened. Scroll should be just for menu not for the whole page.
> 2. Could you check if your fav icon has a cross browser solution? At least I don't see it at iPad for instance.
> 3. Check the image hovers. I see on some breakpoints it moves a little near text blocks. I think, if you will use `transform: scale` it will be fixed.

VladyslavKucher:
"Check the image hovers. I see on some breakpoints it moves a little near text blocks. I think, if you will use transform: scale it will be fixed."

I am using transform: scale in this, but still have this.

polosanya (changes requested):
Almost done! If you have any questions, please ask them in fe_chat and tag mentors. It will be faster because your answers in GitHub will be visible to the mentor only after you re-request review 🙂 
1. Hover effect affects the layout and moves other elements a bit.
Try adding a transparent border for the image's default state and change only its color on hover.  
hover
2. Same here
image
3. Font is not correct here. (Arial instead of Inter)
image
4. Form shouldn't pass the data through the URL on submit
image

tetiana-ivanchenko (approved):

polosanya (changes requested):
Almost perfect! A few small changes left
1. This page doesn't match the design on tablets
image
image
2. It's hard to notice these icons. Check the design. Try to put the image a bit lower
image
image
Same on mobiles
image
3. Burger menu icon is still a bit different. Check border-radius property
image
image

VladyslavKucher:
To change border radius in burger menu, I need to change svg file. If I put border radius manually, everything goes wrong. So, let's leave it how it is right now. I hope it is not a big deal.

polosanya (approved):
Well done! 🔥 
You did a great job!

PR #2312, opened 2023-03-26
landing BOSE, header and menu

diana-tuz:
[DEMO LINK](https://diana-tuz.github.io/layout_miami/)


OleziO (approved):
Hello, everything is fine with the first block, keep working on the next ones :)

OleziO (changes requested):
Just align the picture and text to the center.
image
image

vlad-demchuk (changes requested):
the same
image

nazarmatsevych (approved):
Nice work,
You can also remove scroll when menu is open
image

PR #2328, opened 2023-03-29
Develop

Misha408:
- [DEMO LINK](https://misha408.github.io/layout_miami/)


vlad-demchuk (changes requested):
1. use svg for logos
image

2. make all hovers smoth

OleziO (approved):

PR #2330, opened 2023-03-30
task solution

NasyTheSlayer:
[DEMO LINK](https://NasyTheSlayer.github.io/layout_miami/)


slavanosatskyi (changes requested):
Very nice 😎  just few comments from me

1. Change title of the page, as it is myBike, not Miami.
2. Add favicon
3. Increase line height for your headings according to design
Screenshot 2023-03-30 at 17 55 50

vlad-demchuk (changes requested):
fix the previous comments
image
image

NasyTheSlayer:
Can you elaborate on what i did wrong because i dont understand

PR #2332, opened 2023-03-31
block Header

VasilPan:
[DEMO LINK](https://VasilPan.github.io/layout_miami/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -302,7 +302,7 @@
           <div class="place">
             <div class="place__photos">
               <div class="grid">
-                <div class="grid__item grid__item--desktop-1-8">
+                <div class="grid__item grid__item--desktop-1-3">
                   <div class="place__photo-container">
                     <img
                       src="images/photos/d3.png"
@@ -312,7 +312,7 @@
                   </div>
 
                 </div>
-                <div class="grid__item grid__item--desktop-9-12">
+                <div class="grid__item grid__item--desktop-4-12">
                   <div class="place__photo-container">
                     <img
                       src="images/photos/d3-1.png"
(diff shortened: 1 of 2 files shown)


Volodymyr-Melnyk-Fron10der (changes requested):
Sorry, do not see your DEMO LINK

OleziO (changes requested):
Design: 
image
Demo: 
image

Demo: 
image
Design: 
image

Demo: 
image
Design: 
image

Volodymyr-Melnyk-Fron10der (changes requested):
1. links are not working at all + color should be changed on hover
image

2. check font and fields height + send button shouldn't reload the page, but validate and reset the form
image

3. 
image

4. check it on the design
image
image

5. add favicon
image

6. fix it
image

7. those links lead to the error page
image

8. this link is not working correct
image

IvanFesenko (changes requested):
Form do not work. Add validation
https://user-images.githubusercontent.com/61904995/233379389-4aa608a5-4496-4ab2-9480-0984c126e2d8.mp4

2023-04-20_16h22_36
Check this case. To fix increase z-index for menu

one more case
https://user-images.githubusercontent.com/61904995/233380299-ede69407-91f6-42c1-858e-23c75cf18815.mp4

loralevitska (changes requested):

- make phone number clickable
image
- the cross icon must be on the same level as the burger menu icon
- add smooth scroll
- fix number click, you have syntax error and add [`tel:`](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_link_phoneto)
image

DarkMistyRoom (changes requested):
- Add a smooth scroll for the whole page
- remove outline
image

- check the font. It is different in the textarea
image

- I see no adaptation for the tab screen. Please implement it
- the background moved left too much. Use image directly from the mobile mockup
image

- add a scroll in the menu when the screen has less height than the menu block

image

SanyaBratashchuk (changes requested):
image
 Menu doesn't scroll in lanscape mode

OleziO (changes requested):
Demo: 
image
Design: 
image


Demo: 
image
Design: 
image

nazarmatsevych (approved):
👍

PR #2335, opened 2023-04-01
layout_bose

artem-ivashchenko:
[DEMO LINK](https://artem-ivashchenko.github.io/layout_miami/)


agniya-i (changes requested):
Good job! Keep working)


Make it the same color as icon
Screenshot 2023-04-01 at 19 19 16

OleziO (approved):

PR #2339, opened 2023-04-03
Header without menu

roman-tarasov90:
https://roman-tarasov90.github.io/layout_miami/


etojeDenys (changes requested):
6. Same when you click on logo. There shouldn't be any error.

etojeDenys (changes requested):
1.
image

2. 
image
3. disable scroll when menu is open
image

src/styles/blocks/_header.scss
```diff
@@ -34,8 +34,8 @@
       &::before {
         content: "+1 234 555-55-55";
         position: relative;
-        top: 50%;
```
roman-tarasov90: Why did it stop working?

YegorKorenkov: I think this doesn't work because the element's dimensions are only given by padding(I make position absolute for pseudo before) and this not work for relative elements. I made the pseudo absolute and set the values as in the screenshot
image
image

YegorKorenkov (changes requested):
Looks good, keep working on the next sections

VitaliyBondarenko1982 (changes requested):
Looks good.
Needs some improvements.
- check background position. maybe need to set on top. on wide screens rather hard to reas phone
Screenshot from 2023-04-28 12-19-57
- form should be cleared after submit.
  and should be listener for `submit` event on form
  change your code in main.js on such
  ```
   const form = document.querySelector('form');

    button.addEventListener('submit', buttonClick);

   function buttonClick(event {
     event.preventDefault();
     form.reset();
  }

OleziO (changes requested):
Fix picture. 
image

The menu button is hard to see

witflash (commented):
All looks great to me! 
Well done 👍

witflash (approved):
All looks great to me!
Well done 👍

PR #2342, opened 2023-04-03
menu and header were created

LinaDo1609:
[DEMO LINK](https://LinaDo1609.github.io/layout_miami/) 

I have created just header and menu.
according to instruction I have to create pull request after imposing each blocks


SerhiiKirik (changes requested):
Keep it up 🚀

DarkMistyRoom (changes requested):
Looks good!

- But the close menu button links to 404

image

- these guys should be the same tel links as the phone in the header
image

- I can't scroll the page( Only way to move below the header is menu links
- add hover effects to all images and links
- check the `border-radius` of the `textarea `and its placeholder
image

- all form fields should be required
- don't reload the page on form submitting, just reset the form
- please try to improve the header background on the mobile. Looks like the bike moved left too much. You can take the image directly from the mobile mockup
image

- check the side indents, please. There should be 20px on the mobile
image

- add vertical scroll to the menu to handle a case when the screen is smaller than menu block
image

LinaDo1609:
1. But the close menu button links to 404 - how to check does it have eror 404? i don't understand how you check it
2. about bg of header, i won't change it, how it looks it depend on the height of screen. i have tested different bg (mobile, tablet,desktop) and also its position it doesn't work, if it looks good on mobile, it looks bad on desktop.
3. check the side indents, please. There should be 20px on the mobile - according to mocup, the container should be 320px and have paddings 20px. or i have misunderstanding? container shouldn't be 320px, it just need a paddings 20px? 
if yes, i just change the size of container for mobile from max-width:320px to min-width:320px

LinaDo1609:
i have solved the first point.
the only question is:

check the side indents, please. There should be 20px on the mobile - according to mocup, the container should be 320px and have paddings 20px. or i have misunderstanding? container shouldn't be 320px, it just need a paddings 20px?
if yes, i just change the size of container for mobile from max-width:320px to min-width:320px

tetiana-ivanchenko (changes requested):
GJ, just Change text color on hover for phone, email and address

DarkMistyRoom (changes requested):
Almost there)
- I suggest you use the 'original' image for the desktop and tab, and change it on the mobile
- the container shouldn't affect anything on mobile or tab. The content width should be flexible on these screens but side indents should be hardcoded

- compare these indents. They should be equal

image
image

- make menu closing smooth as well as opening
- make the background flexible. look, it is still desktop but the background moved left
image

- fix the hover effect for the button. something went wrong with it
image

- fix the font in the textarea. It's different than in the other fields
image

- use submit button for the form with this js to make it work properly
`const form = document.querySelector('form');

function submit(event) {
  event.preventDefault();
  form.reset();
}

form.addEventListener('submit', submit);
`

SanyaBratashchuk (changes requested):
1. Make all fields required

TarasHoliuk:
Very nice :fire: 

The main thing that should be fixed:
- centrify bikes there (since it's centrified on the Tablet view, the same should be in the Mobile view. now looks a bit strange on > 320px width screens):
image
image

- also, would be nice to add transition (for color change) here: 
image

TarasHoliuk (changes requested):
You did an amazing job :+1: 
Let's improve a few more things

DarkMistyRoom (approved):
Well done) But fix one last moment:
- there is still another font in the textarea
image

PR #2343, opened 2023-04-04
Task solution

mariuszgit:
Almost finished the page.
Live demo here: https://mariuszgit.github.io/layout_miami/


src/index.html
```diff
@@ -1,13 +1,272 @@
...
+      </div>
+
+      <h1 class="title title--level1 title--decorated header__title">
+        The world shades.
+        <br>
+        Your music shines.
+      </h1>
+    </header>
+
+    <div class="menu" id="menu">
```
Radoslaw-Czerniawski, line 47: All of this could be a part of the header

src/index.html
```diff
@@ -1,13 +1,272 @@
...
-    <title>Miami</title>
+    <title>Bose</title>
     <link rel="stylesheet" href="./styles/main.scss">
     <script src="scripts/main.js" defer></script>
   </head>
   <body>
-    <h1>Miami</h1>
+    <header class="header" id="home">
+      <div class="header-links header__navigation">
+        <a href="index.html" class="header-links__logo">
```
Radoslaw-Czerniawski, line 21: Redirect to '#' rather than index.html

src/index.html
```diff
@@ -1,13 +1,272 @@
...
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>Miami</title>
+    <title>Bose</title>
     <link rel="stylesheet" href="./styles/main.scss">
     <script src="scripts/main.js" defer></script>
   </head>
   <body>
-    <h1>Miami</h1>
+    <header class="header" id="home">
+      <div class="header-links header__navigation">
```
Radoslaw-Czerniawski, line 20: You can have multiple nav blocks on your page and this div could be one

src/index.html
```diff
@@ -1,13 +1,272 @@
...
+          <img
+            src="images/phone-icon.svg"
+            alt="Phone icon"
+            class="phone__icon"
+          >
+        </a>
+
+        <a href="#menu" class="header-links__burger burger">
+          <span class="burger__label">Open menu</span>
+          <!-- <img src="images/ham-sm-icon.svg" class="burger" alt="Hamburger icon"> -->
```
Radoslaw-Czerniawski, line 36: Remove this comment

src/index.html
```diff
@@ -1,13 +1,272 @@
...
+          <span class="phone__label">+1 234 5555-55-55</span>
+          <img
+            src="images/phone-icon.svg"
+            alt="Phone icon"
+            class="phone__icon"
+          >
+        </a>
+
+        <a href="#menu" class="header-links__burger burger">
+          <span class="burger__label">Open menu</span>
```
Radoslaw-Czerniawski, line 35: If you don't display content of this span you may remove it's text, also consider a different tag maybe a link / button?

src/index.html
```diff
@@ -1,13 +1,272 @@
...
+    <main class="main-section">
+      <section class="recommended main-section__section" id="recommended">
+        <h2 class="title title--level2 title--center recommended__title">Recommended</h2>
+
+        <article class="item-box recommended__item">
+          <div class="item-box__photo">
+            <img src="images/photos/image1.png" alt="Photo of Bose portable Smart speaker">
+          </div>
+
+          <h4 class="title title--level4 item-box__name">Bose portable Smart speaker</h4>
```
Radoslaw-Czerniawski, line 89: These could be h3 rather than h4 also don't skip heading levels, if you are not using h3 at all and use h4 just use h3

src/index.html
```diff
@@ -1,13 +1,272 @@
...
+        </article>
+
+        <a href="#" class="btn btn--large product-categories__button">View all</a>
+      </section>
+
+      <section class="why-buy main-section__section" id="how-to-buy">
+        <h2 class="title title--level2 why-buy__title">Why buy direct from Bose</h2>
+
+        <ul class="why-buy__list list">
+          <li><span class="title title--level4">Free 2-day shipping and returns</span></li>
```
Radoslaw-Czerniawski, line 194: Not sure if you semantically need those spans inside li's here, I would personally just put text inside li without those spans

src/index.html
```diff
@@ -1,13 +1,272 @@
...
+          <h4 class="title title--level4 single-category__title">Audio sunglasses</h4>
+        </article>
+
+        <a href="#" class="btn btn--large product-categories__button">View all</a>
+      </section>
+
+      <section class="why-buy main-section__section" id="how-to-buy">
+        <h2 class="title title--level2 why-buy__title">Why buy direct from Bose</h2>
+
+        <ul class="why-buy__list list">
```
Radoslaw-Czerniawski, line 193: This doesn't need to be a separate block, it's tied to the why-buy block and you are not using block element classes inside either way.

Radoslaw-Czerniawski (changes requested):
Generally add some of these changes that I request like using h3, redirects or header structure.I would try to decrease the amount of blocks used here, some of them are unnecessary as you are not using their elements classes (i don't mean use the classes but rather get rid of some block classes). If your block has only one or two css rules then there's not point in declaring it as bem block but rather it being an element of a larger one.
Also you could group your blocks in header, main folders something like this:
image
Overall the layout is very good, it looks and behaves like it was required. Nice job :D

src/styles/blocks/_page-navigation.scss
```diff
@@ -0,0 +1,18 @@
+.nav {
```
Radoslaw-Czerniawski, line 1: This whole file may be removed if the code here is commented

src/styles/blocks/_main-section.scss
```diff
@@ -0,0 +1,15 @@
+.main-section {
+  // display: flex;
```
Radoslaw-Czerniawski, line 2: You may remove those comments

Radoslaw-Czerniawski (approved):
Very good implementation of the layout, some comments from me for the future:
- Try to use class names that are not specifically tied to blocks content, make them generic and descriptive as content of the block may change (for example why-buy's title may change in the future)
- Try not to leave commented code 
- This is based on the structure of our bootcamp but in your future workplace split such giant prs into separate smaller ones or try to make micro commits for example 'add header block', 'add classes for the header block', 'add styling for the header block' -> it's easier for the reviewer to look through pr as commits create a documented history of changes

PR #2344, opened 2023-04-04
Develop

Vlad-Prytula:
[DEMO LINK](https://vlad-prytula.github.io/layout_miami/)


tetiana-ivanchenko (changes requested):
I guess I see just start of burger menu, that impossible to close

DarkMistyRoom (changes requested):
Keep going! You can make next request after finishing whole page

- fix the logo link. Now it leads to 404
image

- center the phone number
image

- close button should be in the same position as the menu button
image

Volodymyr-Melnyk-Fron10der (changes requested):
1. Logo leads to the error
image

2. Add favicon, change the title according to the website theme
image

3. 
image
image
image


4. too much space here 
image
image

5. Form submit should validate and reset the form, not reload the page

6. check the design plz, looks like you didn't pay attention for tablet at all)
image
image
image
image

etojeDenys (changes requested):
1. Add a favicon
2. Same when you click on logo. There shouldn't be any error.
3. Pictures in Gallery should increase on hover
4. Placeholders in the forms suggest what to enter; apply validation of the form fields (required, email / tel etc.), then it is clear in what format to enter the data
5. Form shouldn't be submitted if some of the fields are not filled

PR #2347, opened 2023-04-08
MyBike. Header section

AllaSerhiienko:


ivanchuk-yurii (changes requested):
you should finish all page and only then submit for review. also don't forget about demo link

DarkMistyRoom (changes requested):
Great job! Just a few comments:
- please check this top indent. It's different in the header and in the menu
image
image

- make this button the same as the other one, because now it is always in the hove state) there is an example of hover in the mockup
image

SanyaBratashchuk (approved):

PR #2351, opened 2023-04-11
add task solution

SerhiiVoitiuk:
https://serhiivoitiuk.github.io/layout_miami/


sTorba24 (changes requested):
Good work! But you don't need to create a pr for 'miami' landing - it's for practice
Your task is to implement one of 3 landings from description
image

PR #2358, opened 2023-04-12
Develop

oleksandr-rudiuk:


Volodymyr-Melnyk-Fron10der (changes requested):
1. check this padding plz and move on👍
image

anstsot (changes requested):
Good start, keep going! Re-request the review once you finish the page

loralevitska (changes requested):
- add hover effect to images in "Browse Bose products by category" block
- add favicon
- it would be good if the "View all" button led somewhere, for example, to their real site, which can be found on the Internet
- always open links in new tab
- add hover effect to contacts in "Contact us" block
- don't reload page after form submit
- fix cursor in message input it should start at the beginning and not in the middle of the line
- disable scroll while menu is open
- fix indents according design on mobile mode
image

oleksandr-rudiuk:
1. fix indents according design on mobile mode. Distance between pictures - 20px when display width 320px. As used grid, when display stretched, stretched pictures and distance between them. 2. add cursor pointer. For icon "menu" I used "mice-cursor" according Figma sample.

oleksandr-rudiuk:
" these indents should be the same according to the mockup". According Figma, text start point sharply under left side of picture. At display width 320px, indents 20px. When display stretch, picture size stays the same, so indent increase.

SanyaBratashchuk (changes requested):
1. make transition for hover effects work in both directions
image
3. All external links should open in the separate tab 
4. Can't scroll menu in the landscape mode 
image

src/index.html
```diff
@@ -1,13 +1,315 @@
...
+        <div class="header__top">
+          <a href="#" class="logo">
+            <img 
+              class="logo_image"
+              src="./images/bose/logo.svg"
+              alt="Bose logo"
+            >
+          </a>
+          <a
+            href="+1 234 555-55-55"
```
ivanchuk-yurii, line 31:
```suggestion
            href="tel:+12345555555"
```
fix everywhere

ivanchuk-yurii (changes requested):
points 2, 5 from previous review haven't been fixed. it's worth downloading images in higher resolution so when you scale they look clear

oleksandr-rudiuk:
1. Скролл з'являється за висоти екрану 340рх. За ширини екрану більше 1100рх він блокується js кодом.
2. Form submit shouldn't trigger page refresh Повинно блокуватись  js кодом, наданим [SanyaBratashchuk](https://github.com/SanyaBratashchuk)

Anastasiia-Svintsova (changes requested):
- These 2 sections should have the same width on the desktop. Check the Figma design, please 
image
image

- Check the menu icon size 
image
image

- The design doesn’t have the `Tel:` before the phone number 
image

- The hover effect for the links doesn't look good. Blue is not in the color scheme of the page. it's better just to make the text darker. And the phone link doesn't work 

https://user-images.githubusercontent.com/99403198/233971936-7d68a074-177f-454a-b3c7-7e9ff1bfb9c1.mov

- The page shouldn't reload after the form submission. And the user shouldn’t see the info from the form in the URL.

https://user-images.githubusercontent.com/99403198/233971909-0a57aad4-9c94-43ad-be23-21fa25f54eae.mov

qoa11a (approved):

PR #2370, opened 2023-04-18
Develop

VolodymyrPst:
[DEMO LINK](https://VolodymyrPst.github.io/layout_miami/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,7 +7,7 @@
     <link
       rel="icon"
       type="icon"
-      href="./images/logo.svg"
+      href="./images/favicon.ico"
     >
     <link rel="stylesheet" href="./styles/main.scss">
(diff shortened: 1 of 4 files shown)


Viktor-Kost (changes requested):
Hello!)
Good job!
Some moments to fix: 
1) I would recommend to change your favicon to make it more readable 
Screenshot 2023-04-18 at 18 31 05

2) when you click on logo. There shouldn't be any error.
3) pictures in Gallery should increase on hover
4) page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)
5) form shouldn't be submitted if some of the fields are not filled
6) change text color on hover for phone, email and address
7) when you click on external links like address it should be opened in a new tab

PR #2372, opened 2023-04-19
Bose

levchenko-dmytro:
- [DEMO LINK](https://levchenko-dmytro.github.io/layout_miami/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -335,12 +335,14 @@
                   name="name"
                   class="contact-us__field form__field"
                   placeholder="Name"
+                  required
                 >
                 <input
                   type="email"
                   name="email"
                   class="contact-us__field form__field"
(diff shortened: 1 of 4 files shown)


Volodymyr-Melnyk-Fron10der (changes requested):
Keep going

YegorKorenkov (changes requested):
One minor mistake
1. each navigation element should lead to the corresponding section
image

Volodymyr-Melnyk-Fron10der (changes requested):
1. Form shouldn't reload the page. just validate th inputs and reset them. And fix the border radius
image

2. implement highlighted points
image

nazarmatsevych (approved):
Good work,

just small suggestion:

It's better to make form fields required
image

PR #2375, opened 2023-04-20
Added MyBike

PavloSich:
[DEMO LINK](https://pavlosich.github.io/layout_miami/)


loralevitska (changes requested):

- add favicon
- fix click on logo (now we have 404 error)
- fix this according design and make number clickable
image
- add hover effect to images in `The Details` block
- download image without any elements
image
- always open links in new tab
- fix form submit (now it's 404 error) it shouldn't reload page too
- clear form after submit
- fix border radius for textarea according design 
- the cross icon must be on the same level as the burger menu icon
- disable scroll while menu is open

SanyaBratashchuk (changes requested):
1. external links should be open in the separate tab 
2. make all fields required
3. Menu is not scrollable in landscape mode 
image

OleziO (changes requested):
Fix the image position. 
image

SanyaBratashchuk (approved):

PR #2384, opened 2023-04-24
MyBike landing

olena-hudimenko:
[DEMO LINK](https://olena-hudimenko.github.io/layout_miami/)

I have not yet been able to solve the problem of increasing the images in the gallery, in the process.


IvanFesenko (changes requested):
Check this case. To fix increase z-index for menu
2023-04-24_15h06_19

Also add some hocer effect to links
2023-04-24_15h06_40

olena-hudimenko:
I don't have a problem with the menu scroll and I asked about it in Slack, Lora checked the link, and there is no this problem

loralevitska (changes requested):

- add cursor pointer to all clickable elements (logo, menu, buttons)
- fix border radius for textarea in form according design

olena-hudimenko:
I see cursors pointer on all clickable elements, should I write it explicitly?

qoa11a (approved):

PR #2385, opened 2023-04-24
my-bike landing 

DidenkoDenis306:
[DEMO LINK](https://DidenkoDenis306.github.io/layout_miami/)


Ihor-Karpyn (changes requested):
- Decrease animation time of block appears, animation should not cry "look i am here". It should be light as posible
- The same with phone number animation, it's looks like advertising banner from 2000
- The same with bicycle price
- About button go up. Let's change icon to somethimg different. Some white filled arrow without backgroud around
- I can't submit the form
- All links on external source should open in new tab
- Menu on mobile 

image

Ihor-Karpyn (approved):
🔥

PR #2388, opened 2023-04-25
Ulcja/25.04.23

Ulcja:
[DEMO LINK](https://Ulcja.github.io/layout_miami/)


anstsot (changes requested):
Good start, keep going! Re-request the review once you finish the whole page

loralevitska (changes requested):

- fix 404 error by click on logo
image
- set right padding for menu 
image
- check the cross icon in menu and burger menu icon on page, it should be on the same position and have the same width & height
- make this phone numbers clickable with `tel:` links
image
- disable scroll when menu is open. Check 18 point in [checklist](https://github.com/mate-academy/layout_miami/blob/master/checklist.md)
image
- add hover effects for images on page
- avoid fields jumping by hovering in form. You may add transparent border and then only change it color by hover 
- make all field required in form
- don't scroll page to top by form submit. Just send info from form and clear it
- add hover effect and make it links
image
- add underline to navigation links by hovering on it according design
image
- make it links 
image
- remove this `X` 
image

PR #2389, opened 2023-04-25
The MET

generatedman:
[DEMO LINK](https://generatedman.github.io/layout_miami/)


Ihor-Karpyn (approved):
It's no need to show some alert, just make send-button type submit

image

PR #2395, opened 2023-04-26
Develop

AlisaSkryp:
https://AlisaSkryp.github.io/layout_miami/


Viktor-Kost (changes requested):
Hello!)
Keep going!
Some moments to fix: 
1) change phone number positioning
Screenshot 2023-04-26 at 18 37 20

2) phone number at menu should be clickable at give a chance to call

PR #2398, opened 2023-04-26
Develop

VolodymyrPst:
https://volodymyrpst.github.io/layout_miami/


vlad-demchuk (approved):
Looks good

add transition
image

PR #2401, opened 2023-04-28
Add header

zuzu6:
[DEMO LINK](https://zuzu6.github.io/layout_miami/)


tetiana-ivanchenko (changes requested):
shouldn't be an error when click on the logo
request review when your layout will beready

PR #2402, opened 2023-04-28
add task solution

vovaAlieksieiev:
[DEMO LINK](https://vovaAlieksieiev.github.io/layout_miami/)


nazarmatsevych (changes requested):
1. images should not overlap each other
image

2. fix background image (tablet and mofile view) so that it looks similar to the design
image

vlad-demchuk (changes requested):
not fixed 
> images should not overlap each other
image
the same with other comments

make sure you pushed your changes

etojeDenys (approved):
awesome

PR #2409, opened 2023-05-01
add mobile header

annholovko:
[DEMO LINK](https://annholovko.github.io/layout_miami/)


Anna-Tkachenko (changes requested):
Just one moment:
1) Page shouldn't be reloaded on form submitting

PR #2412, opened 2023-05-02
Bike landing

okosohor:
[DEMO LINK](https://okosohor.github.io/layout_miami/)


33Orange (changes requested):
- If you click logo it will give you error.
- You should have added hover effects to phone, email and address.
- In contact us section link don't have `target="_blank"` attribute. (`Location-related addresses / links should open google maps in a new tab target="_blank"`).
- You should have added hover effects to images (`Pictures in Gallery should increase on hover`).
- Now if you click on the submit form, the page will reload. You should have prevented this. (e.preventDefault() in submit function).

Everything else is done fine, but bug fixes are needed (by checklist).

33Orange (changes requested):
Almost done!

You forgot to add hover to link in contact section. You can change color on hover
`You should have added hover effects to phone, email and address.`

And my comment about `<title>`. That was my mistake. You need to use tags like `h1`, `h2` for titles.
`<title>` is element for [header](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title).

Please request a review after fixes

33Orange (approved):
Good job

PR #2414, opened 2023-05-02
MyBike Landing page

serentinos:
[DEMO LINK](https://serentinos.github.io/layout_miami/)


33Orange (changes requested):
- You should have added hover effects to phone, email and address.
- In contact us section link don't have target="_blank" attribute. (Location-related addresses / links should open google maps in a new tab target="_blank").

Everything else is done well, but you need to fix the errors.

P.s. Cool animation ! 😎

Anna-Tkachenko (changes requested):
Good job!
Some comments:
1) Add some styles on hover state for the phone link in the opened menu
2) I have an access for other elements while menu is opened with the `tab` button, will be great to hide them. [link](https://prnt.sc/9Gca_RRjJ8D1)
3) Make form fields the required ones
4) Form fields should be cleaned up after form submit

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #2415, opened 2023-05-02
add desktop and mobile header

opiquem:
[DEMO LINK](https://maksym-bobryk.github.io/layout_miami/)


Ihor-Karpyn (approved):
Gj, left fes suggestions

- links to external resources should open in a new tab
image

- the form is not submitted, there is no validation
image

- let's adapt menu
image

PR #2416, opened 2023-05-02
landing header(desktop/mobile)

sasdaiv:
[DEMO LINK](https://sasdaiv.github.io/layout_miami/)


etojeDenys (changes requested):
"Same when you click on logo. There shouldn't be any error."

keep going

PR #2418, opened 2023-05-02
layout_miami

bohdan-mykhailenko:
[DEMO LINK](https://bohdan-mykhailenko.github.io/layout_miami/)


silchyn (changes requested):
- Scroll is not smooth
- Is this border around images on hover expected?
- Form is not cleared on submit
- Why would I submit the form without a message?
- Why would I need these query parameters?

silchyn (changes requested):
- Something wrong with font-family
image
image
image

33Orange (approved):
All comments was fixed.
Well done 🔥

naziamoff (approved):

PR #2419, opened 2023-05-02
bose landing 

cephy77:
- [DEMO LINK](https://cephy77.github.io/layout_miami/)


33Orange (changes requested):
- If you click logo it will give you error.
- In contact us section link don't have target="_blank" attribute. (`Location-related addresses / links should open google maps in a new tab target="_blank"`).
- You should have added hover effects to images (`Pictures in Gallery should increase on hover`).
- You should have added require attribute to the inputs in form. Now all fields are optional and the form can be submitted with an empty field. (`apply validation of the form fields (required, email / tel etc.), then it is clear in what format to enter the data`, `Form shouldn't be submitted if some of the fields are not filled`).

silchyn (changes requested):
Uploading Loom Message - 8 May 2023 (1).mp4…

silchyn (changes requested):
https://user-images.githubusercontent.com/30828805/236828636-4455f7fe-ca04-4a7e-bfe9-6bb7885f19bf.mp4

silchyn (changes requested):
- There is a blink between the phone icon and the phone number in the header
- Error when clicking on the logo in the menu
- Links should be opened in a new tab
- Maybe we should scroll the user to the top instead of that alert
- Header takes the whole width on large displays

silchyn (approved):

PR #2421, opened 2023-05-03
header and hero-section

skinnyjazz:
[DEMO LINK](https://skinnyjazz.github.io/layout_miami/)


33Orange (changes requested):
- Clicking to phone number in "contact us" section doesn't work. (`When you click on phone icon or phone number in contacts section, make sure that there is no 404 error, make it a real link to start a call on device`)
- Clicking to address in "contact us" section doesn't work. (`When clicking on any location / address - prevent errors and make it to open location in Google Maps`). Don't forget to add target="_blank" attribute :).
- You should have added require attribute to the inputs in form. Now all fields are optional and the form can be submitted with an empty field. (`apply validation of the form fields (required, email / tel etc.), then it is clear in what format to enter the data, Form shouldn't be submitted if some of the fields are not filled`).

dinaminahi (changes requested):
Please, check p.14 from checklist 
`Message` is also required

33Orange (approved):
Good job

PR #2431, opened 2023-05-03
add header and menu

amashutikov:
- [DEMO LINK](https://amashutikov.github.io/layout_miami/)


silchyn (changes requested):
- Why would I submit the form with an empty message?
- The URL has query params after form submission, but what is their purpose?
- The footer logo is in the left bottom corner for a large display

33Orange (approved):
Well done ! 🔥 

All comments was fixed.

PR #2433, opened 2023-05-03
created header

RoksolanaTylipska:
[DEMO LINK](https://RoksolanaTylipska.github.io/layout_miami/)


33Orange (changes requested):
- It would be better to use some transitions when you have effects on hover. (Header: phone, burger menu. Footer: phone, email, address). 
- If you want to show that the element is interactive would be better to use `cursor: pointer`
- When the menu is open I still can scroll web page.
- The Z-index of the menu should be higher than all other elements. Now pictures from the gallery go on top of the menu
- Form should be cleared after submission

RoksolanaTylipska:
> * If you want to show that the element is interactive would be better to use `cursor: pointer`

image
Hello, please, pay attention - this black arrow on `a` tags is `cursor: pointer`

33Orange (changes requested):
You didn't fix all my previous comments:
1. I have an access for other elements while menu is opened with the `tab` button, will be great to hide them. `The Z-index of the menu should be higher than all other elements. Now pictures from the gallery go on top of the menu`
image

Everything else was fixed.

p.s. I saw that you using custom cursor. It's not best practice to change the cursor for interaction. If you make a custom cursor, then it must be custom on the entire site, and not just on the hover. Now it is not very clear what kind of arrow it is and what it is for.

dinaminahi (changes requested):
can't access demo link 
image

RoksolanaTylipska:
> can't access demo link image

Hello! Please check again, everything should be OK.

silchyn (changes requested):
- Menu logo gives an error on click
- Check how those should look
image

RoksolanaTylipska:
> * Menu logo gives an error on click
> * Check how those should look
> 
> image

@silchyn  Done.

dinaminahi (approved):
👍

PR #2434, opened 2023-05-04
Met

KatyaZaremska:
[DEMO LINK](https://KatyaZaremska.github.io/layout_miami/)


naziamoff (approved):
Impressive job 🔥 🔥 🔥 
One small thing:

- it is difficult to see the number here, I would think of making the color more contrast

image

Also I have one small suggestion (not required). You can try to use the [aos](https://www.npmjs.com/package/aos) lib for making your page look dynamic

PR #2435, opened 2023-05-04
create header and optimized header for all screen

GlibVK:
Pleas, check the header. This is first block in this task.
https://glibvk.github.io/layout_bike-landing-/


tetiana-ivanchenko (changes requested):
add demo link

anstsot (changes requested):
Good start, keep going! Re-request the review once you're done

PR #2440, opened 2023-05-04
90% of header ready

olehmahera:


33Orange (changes requested):
You need to provide [DEMO LINK](https://%3Cyour_account%3E.github.io/layout_miami/)

PR #2441, opened 2023-05-04
add task solution

SofiaLiubchenko:
[DEMO LINK](https://SofiaLiubchenko.github.io/layout_miami/)


Anna-Tkachenko (changes requested):
Good job!
Some comments:
1) Scroll should be disabled when menu is opened
2) Also check access to the other sections via `tab` button. Will be great to hide elements when menu is opened
3) Make form fields required. For now as user I try to submit an empty form and don't see any errors
4) Form fields should be cleaned up after form submitting
5) Check the logo existence in footer in figma. Looks like it shouldn't be there

33Orange (changes requested):
The demo version of the site still has all the bugs that @Anna-Tkachenko listed.
Maybe you haven't updated the demo ?

33Orange (changes requested):
1. Form must validate email field. Right now I can write anything to email input and send this form.
2. Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)
3. Links in footer must have some hover effect (change color for example)

33Orange (approved):
Good job.

For the future.
HTML validation is much more flexible and understandable than the JS alert window. You also need to know how to work with her.

PR #2444, opened 2023-05-04
add header without menu

growingananas:
[DEMO LINK](https://growingananas.github.io/layout_miami/)


silchyn (changes requested):
- Page is not scrollable initially
- Logo gives error on click
- Ensure links do not lead to errors
- Form submission leads to error
- Why would I submit the form without a message?

silchyn (changes requested):
- Menu logo gives an error on click
- The URL contains query params after form submission; what is their purpose?
- Content stretches for the whole width, even on large displays, which is a lot of pixels
- Phone number in header overlaps logo on a small display

silchyn (approved):

PR #2448, opened 2023-05-04
BOSE landing

PANCH1811:
[DEMO LINK](https://panch1811.github.io/layout_miami/)


naziamoff (changes requested):
I love it. Still requires some polishing

- doublecheck the icons with figma. I might be wrong, but they seem too small to me
- great job with making the page dynamic 🔥. However, it looks kinda too dynamic and abrupt. I suggest you to think of sth to make it move more natural
- should we have the data I've submitted the form with in the address line?

image

silchyn (changes requested):
- Scroll is not smooth
- Form submission gives error
- Why would I submit form with empty message?

silchyn:
> @naziamoff form method fixed, icons exactly like in figma, all 24x24. What u mean "think of sth"?

For example add transition where you don't have it

PANCH1811:
> 

U can't submit an empty form
image

PANCH1811:
@silchyn 

- Scroll is smooth now
- Links in `Contact Us` becomes bold on hover.
- Form submission gives a `Thank you` message
- Now the text area in form is required

and a few transitions added.

silchyn (approved):
Maps should open in a new tab!

naziamoff (approved):

PR #2449, opened 2023-05-04
Develop

vitvadym:
[DEMO LINK](https://vitvadym.github.io/layout_miami/)


silchyn (changes requested):
- The form button has a border
- I can submit an empty form
- Why would I need these query parameters?
- I can't close the menu on a small display
- Check the `Explore` button on a large display

dinaminahi (changes requested):
I have horizontal scroll on small devices
image

33Orange (changes requested):
- I still have horizontal scroll
image
- I have spaces in `textarea` by default
image
- The form fields must be `required`. Now I can submit a form with empty message field. 
`[Checklist p14: Form shouldn't be submitted if some of the fields are not filled]`
- The menu should go on top of everything on the site. If the menu is open and I press `tab`, then I can see the content.
image

PR #2453, opened 2023-05-04
added header block

dimoshen:
[DEMO LINK](https://dimoshen.github.io/layout_miami/)

still need to add an element under h1 and correctly scale the background image.


33Orange (changes requested):
- You need to add favicon!
- It would be better to use some transitions when you have effects on hover. (Header: phone, burger menu. Footer: phone, email, address).
- In contact us section link don't have target="_blank" attribute. (`Location-related addresses / links should open google maps in a new tab target="_blank"`).
- Pictures in Gallery should increase on hover. 
- In mobile version you need to move background image because the buttons aren't visible. 
image

- Form should be cleared after submission

polosanya (changes requested):
Comment above

PR #2454, opened 2023-05-05
add adaptive header and menu

PanyaPrimal:
[DEMO LINK](https://PanyaPrimal.github.io/layout_miami/)


silchyn (changes requested):
- Form does not seem to work
- Check `Contact Us` on a small display

33Orange (changes requested):
Almost done!

A few moments:
1. I can send form without message.
image

2. You need to add hover only for link. To make it clear that this is a link and there is an interactive. And add transitions for hover effect.

image

naziamoff (approved):
Good job! One small thing left

- different fonts on different form fields
image

PR #2458, opened 2023-05-05
add header

maryanmats:
[DEMO LINK](https://maryanmats.github.io/layout_miami/)


silchyn (changes requested):
- Check menu links on the hover
- Map should open in a new tab
- Form is broken

silchyn (changes requested):
Check DEMO link

dinaminahi (changes requested):
Few more things to do: 

1.  Checklist p. 4  - You need to define hover styles for links in the contact-us section. You can choose any color that matches your design.

2. Checklist p. 15 - Page shouldn`t reload on submitting form

dinaminahi (changes requested):
Okay, the form now isn't reloading the page, but it is not reseting the fields, you can do it like this:
`
e.target.reset()
`

dinaminahi (approved):
👍

PR #2459, opened 2023-05-05
added landing

holytimy4:
[DEMO LINK](https://holytimy4.github.io/layout_miami/)


silchyn (changes requested):
- Buttons have a border
- Why would I submit the form without a message
- Maybe we should scroll the user to the top instead of showing alert

silchyn (changes requested):
Almost done!

- Phone number in header overlaps logo for small display
image
image

silchyn (changes requested):
- Bikes are cropped on hover. Is it expected?
image
image

silchyn (changes requested):
- Now, it has a horizontal scroll

33Orange (approved):
A few moments:
1. Hero section not even close to design on mobile

image

2. On a mobile device, all text in sections overlaps each other during in-view animation and it doesn't look nice

PR #2483, opened 2023-05-17
MyBike layout

ShamievDavid:
[DEMO LINK](https://ShamievDavid.github.io/layout_miami/)


IvanFesenko (changes requested):
Well done - keep going

etojeDenys (changes requested):
fix this mismatch:
image
image
2. and add hover to photos

ShamievDavid:
For now i added 'details block' without images. Because I didn't find solution, so i will go to Q&A with that question tommorow.

mykhalenych (changes requested):
https://github.com/mate-academy/layout_miami/assets/53228259/0e265b29-87ee-49a4-bdfa-8f32a2fee307

1. fix scroll issue

Screenshot 2023-05-22 at 16 11 50
2. add hover to links
3. add validation to inputs

ShamievDavid:
1. i fixed scroll issue 
2. i added on hover (there is no design, so i just made color "yellow" on hover. 
3. i don't really understood, what do you mean by validation in this case? 
I made every field "required" but previously i turn off default behavior of the form, because I saw it in checklist. 
so now when you click the submit button - happens nothing. Maybe i should use event.preventDefault in another way? 
Or what do you mean by validation? Thank you.

VitaliyBondarenko1982 (changes requested):
Almost done.
Form still reload the page 
Screenshot from 2023-05-23 15-44-55
To fix it add such code to main.js

```
  const form = document.querySelector('form');

form.addEventListener('submit', (event) => {
  event.preventDefault();

  form.reset();
})

VitaliyBondarenko1982 (approved):
Looks good to me.

VitaliyBondarenko1982 (approved):
Looks good to me.

VitaliyBondarenko1982 (approved):

PR #2486, opened 2023-05-17
MyBike landing

stnayn:
[DEMO LINK](https://stnayn.github.io/layout_miami/)


TarasHoliuk:
1. In the mockup this section looks a bit different in mobile (and tablet as well) version

App:

image

Mockup (images have square shape):

image

2. Use the mobile (and tablet) version image from the mockup in the header:

Mockup:
image

App:
image

3. There shouldn't be an icon in the corner:
image
image


[Checklist](https://github.com/mate-academy/layout_miami/blob/master/checklist.md) points 9, 15 (also, all fields should be required, and reset form on submit - ask in chat if you need help to implement it)

TarasHoliuk (changes requested):
Very good job :+1: 
But let's improve it a bit (comments)

sTorba24 (changes requested):
Looks great! One thing to fix

You need to stop reloading page when form submitted and also would be good to clear it

add js with such code

```
const form = document.getElementById('form')        #add id="form" to form

form.addEventListener('submit', (e) => {
  e.preventDefault();
  form.reset()
})

sTorba24 (approved):
Good work!

PR #2492, opened 2023-05-18
header/footer

MaksymKos:
[DEMO LINK](https://maksymkos.github.io/layout_portfolio/)


polosanya (changes requested):
Nice start!
1. Let's add a favicon for page
image
2. Burger menu doesn't match the design. Also, it should overlay the main screen, not be a separate section
image

etojeDenys (changes requested):
1. When clicking on any location / address - prevent errors and make it to open location in Google Maps
2. Pictures in Gallery should increase on hover
3. Location-related addresses / links should open google maps in a new tab target="_blank"
4.  Apply :hover effect for images on page (testimonials / gallery, other sections).
5.  Placeholders in the forms suggest what to enter; apply validation of the form fields (required, email / tel etc.), then it is clear in what format to enter the data
6. Form shouldn't be submitted if some of the fields are not filled
7. Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)
8. Add a smooth scroll for the whole page

VitaliyBondarenko1982 (changes requested):
Almost done.
Needs some improvements.
1. add transition on hover.
    for example in gallery and check other elements with hover effect
    and check such overlapping after hover.
    
Screenshot from 2023-05-20 17-40-54

    to fix it add overflow: hidden for image container. if image have some margins - better to set them to other elements instead
2. Change text color on hover for phone, email and address ([checklist](https://github.com/mate-academy/layout_miami/blob/master/checklist.md) - point 4)
Screenshot from 2023-05-20 17-41-53
3. submit form still reload the page. and it should be listener for 'submit' event instead 'click'
   also clear form after submit `form.reset()`

etojeDenys (approved):
awesome

PR #2493, opened 2023-05-18
Develop

BOHDANLIUYUI:
[DEMO LINK](https://BOHDANLIUYUI.github.io/layout_miami/)


tetiana-ivanchenko (changes requested):
request review when your header will be ready, and your design should match with the mockup

PR #2495, opened 2023-05-19
boss

MariiaSuper:
https://MariiaSuper.github.io/layout_miami/


nazarmatsevych (changes requested):
1. Remove scroll when menu is open
image

2. Fix text in placeholders due to mockup
image
image

3.  Page should not reloads on submit, also clear form fields on button click
image

4. Add some hover effects to all interactive elements
image

vlad-demchuk (changes requested):
add favicon
image

vlad-demchuk (approved):

PR #2509, opened 2023-05-29
Develop

Solovei-Andrii:
[DEMO LINK](https://Solovei-Andrii.github.io/layout_miami/)


nazarmatsevych (changes requested):
1. It's better to add cursor: pointer to all links and buttons
2. Add hover effects to all interactive elements(links, buttons and so on) and images in gallery
image
image
image
image
3. There are should be 3 links with hover effects
image

4. Form should not reloads on submit, also clear form fields on button click. Make all form fields required
image
5. Add transparent border bottom to list items by default and change only color on hover
image
6. Remove scroll when menu is open
image

PR #2511, opened 2023-05-30
add task solution

ydashko:
seems like on the example but some parts are not working with demo


ydashko:
https://ydashko.github.io/layout_miami/

PR #2528, opened 2023-06-20
Solution header + menu

ArtemVelychko:


lerastarynets (changes requested):
Pls watch the video review :)

https://github.com/mate-academy/layout_miami/assets/77551820/ccf07440-32ab-4805-acc6-2c82f206de18

polosanya (changes requested):
Almost perfect!
1. There are two different phone numbers, but the link is the same. Would be great to fix it
image
image
2. Same in burger menu
image
3. Let's add hover effects to these links too
image
4. The face is covered with the title on mobile and tablet versions. Not critical, but would be great to fix
image
image

polosanya (approved):
Great job!

PR #2533, opened 2023-06-26
Bose Landing

Kacper-Leman:
[DEMO LINK](https://Kacper-Leman.github.io/layout_miami/)


src/index.html
```diff
@@ -1,13 +1,404 @@
...
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" >
+    <link href="https://fonts.googleapis.com/css2?family=Inter:wght@400;700&display=swap" rel="stylesheet">
     <link rel="stylesheet" href="./styles/main.scss">
+
     <script src="scripts/main.js" defer></script>
   </head>
-  <body>
-    <h1>Miami</h1>
+
```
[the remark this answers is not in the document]
Radoslaw-Czerniawski: Generally you should empty lines between siblings:
```html
<body>
  <div>
    <span>1</span>
  
    <p>2</p>
  </div>

  <div>
    <div>
      <button></button>  
    </div>

    <span></span>
  </div>
</body>
```

But there are exceptions for example when you have a bunch of repeated tags in container like here:
```html
<ul>
  <li><a></a></li>
  <li><a></a></li>
  <li><a></a></li>
  <li><a></a></li>
</ul>

<ul>
  <li></li>
  <li></li>
  <li></li>
  <li></li>
</ul>
```

src/index.html
```diff
@@ -1,13 +1,404 @@
...
+  <body class="page__body">
+    <header class="header">
+      <div class="container">
```
Radoslaw-Czerniawski, line 17: One of these divs is not necessary, get rid of it

src/index.html
```diff
@@ -1,13 +1,404 @@
...
+                src="/images/logo.svg"
+                alt="logo"
+              >
+            </a>
+
+            <a
+            href="tel:+1 234 555-55-55"
+            class="icon icon--call"
+            >
+              <div class="icon__tooltip">
```
Radoslaw-Czerniawski, line 32: Don't use divs for everything, this could be span / p tag

src/index.html
```diff
@@ -1,13 +1,404 @@
...
+          >
+            call to order
+          </a>
+        </div>
+      </div>
+    </nav>
+
+    <main>
+      <section class="page__section recommended" id="recommended">
+        <div class="container">
```
Radoslaw-Czerniawski, line 108: This div is unnecessary, you could style the section above instead of adding another wrapper, remove it

src/index.html
```diff
@@ -1,13 +1,404 @@
...
+          <h2 class="page__section-title">
+            Recommended
+          </h2>
+
+          <div class="grid grid--tablet grid--desktop">
```
Radoslaw-Czerniawski, line 113: Same here, there are 3 lvls of unnecessary div nesting, it should look like so:
```html
<article class="cart">
  <div class="cart__photo-container">
    <img
      src="./images/cart-1.png"
      alt="cart-1"
      class="cart__photo"
    >
  </div>

  <h3 class="cart__name">Bose portable Smart speaker</h3>

  <p class="cart__subtitle">Smart home</p>
  <p class="cart__price">$ 399.00</p>
</article>
```

Radoslaw-Czerniawski: Also take notice I've change the div with text into `p` tag, and the wrapper into an [article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article), don't use divs for everything.

Radoslaw-Czerniawski (changes requested):
Generally the layout looks very good, however the code needs some refactoring. 
- Remove all unnecessary wrappers (you have too many nested divs)
- Use more semantic html (for text span / p etc.)
- Based on the comments again look at your indent and empty lines, try to correct it

There are also errors with the layout implementation:

- How the number is positioned differs from the figma design, which wouldn't be that bad if the number was visible:
#### Implementation
image

#### Design
image

---

- Some of the borders shouldn't have (or should have smaller) radius:
#### Implementation
image

#### Design
image

---

- This is not mentioned in the figma design however phone number is also almost invisible on the 320px width, as this project will be added to your portfolio it's best to also correct it:

image

---

Again take a look at the [checklist](https://github.com/mate-academy/layout_miami/blob/master/checklist.md) and ensure everything is implemented according to the requirements.

src/index.html
```diff
@@ -1,13 +1,388 @@
...
+          </div>
+        </div>
+      </section>
+
+      <section class="page__section gallery container" id="categories">
+        <h2 class="page__section-title">
+          Browse Bose <br> products by category
+        </h2>
+
+        <div class="gallery__place place">
```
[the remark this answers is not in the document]
Radoslaw-Czerniawski: - naming of these classes is too vague, by reading this I have no idea what `place` refers to
- alt attributes are supposed to described what resides on the photos, all of them need to be corrected in this pr
  - screen readers / google's robots will only see the alt attribute and based on how well it reflects content on photo (or it's purpose) they will position your website higher in search results, 
  - an example of a good usage of alt attribute for the bose logo in the top right corner `alt="Bose brand logo"`

src/styles/utils/_vars.scss
```diff
@@ -1 +1,3 @@
-$c-gray: #eee;
+$background-color: #fff;
```
Radoslaw-Czerniawski, line 1: - all of the colors used should be put into variables
- don't name color variables with names like `main-color` but rather describe how the color looks like, for example `light-blue`

Radoslaw-Czerniawski (changes requested):
- There is an invisible tag that is possible for me to click on:
image
It redirects me to a non-existing page with 404 error
image
It needs to be fixed

- Generally there are still massive amounts of nested divs
  - reduce their amount to only necessary -> 1, max 2 per group
  - try to use semantic tags for the purpose of grouping, for example don't unnecessarily wrap from or article with divs, those tags alone should be enough
- rest in the comments

src/styles/blocks/icon.scss
```diff
@@ -0,0 +1,83 @@
...
+    letter-spacing: 2px;
+    text-transform: uppercase;
+
+    opacity: 0;
+    pointer-events: none;
+
+    transition: opacity 0.3s;
+  }
+
+  &:hover &__tooltip {
```
Radoslaw-Czerniawski, line 36: You could move this inside the tooltip styles above:
```scss
&__tooltip {
  ...
  opacity: 0;
  pointer-events: none;

  transition: opacity 0.3s;

  &:hover { 
    ... 
  } 
}
```

src/styles/blocks/button.scss
```diff
@@ -0,0 +1,35 @@
+.button {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  width: 100%;
+  height: 56px;
+  background-color: #292929;
```
Radoslaw-Czerniawski, line 8: This color is heavily reused in your stylings, it could be a variable (generally all your colors ideally would be variables)

Radoslaw-Czerniawski (changes requested):
Almost there, some few more changes needed:

- menu on screen size ~1000px and smaller has no space between the edge of the viewport
image

image
<br>
<br>

- I noticed you are not using any kind of reset.scss, you could omit a lot of unnecessary by-hand resets of margins etc. by implementing it (however now it may break some things, it would be best to use it at the beggining of such project)
- left some more comments, generally most important thing now is getting that menu up to shape

src/styles/utils/_reset.scss
```diff
@@ -0,0 +1,7 @@
+body {
```
Radoslaw-Czerniawski, line 1: Resetting is setting default browser styles like for example margins to value of 0, here you only did that for the body. Generally it would be best to use something like the [Meyer Reset](https://meyerweb.com/eric/tools/css/reset/)

Radoslaw-Czerniawski (approved):
There are still some small mistakes, however the layouts looks very good so we can move forward :D

Kacper-Leman:
I willl in free time a little bit fix it becasue I also see few things what wanna change :D

PR #2541, opened 2023-07-11
Develop

HaniaNassalska:
[DEMO LINK](https://hanianassalska.github.io/layout_miami/)


JoyCoffeeAddict (changes requested):
1. You should be able to click on phone number and make a call 
image

2. This animation could be smoother.
image

3. You can read about and use `scroll-behavior` to have better scrolling. 
4. There are some resolutions where you could center this button, try to decide on your own if you want to do it. 
image

5. This would look better on smaller devices (title over description): 
image

6. Reduce the text width on larger devices
image

choeqq (changes requested):
1. Those margins with different background-color looks little off - I would try to stretch hero image little more, to take the whole width and definitely change background to match all other sections, please give menu little more space on bigger screens, it looks really squeezed
image
2. Please give nav button transition or omit those animations on hover
3. Address should take us not only to Google Maps but to precise location 
4. Form should not reloard our page [checklist ](https://github.com/mate-academy/layout_miami/blob/master/checklist.md)point 15

Most importantly please fix margin around content - content should take as much space as possible - it's most important 😄

src/index.html
```diff
@@ -3,11 +3,183 @@
...
-    <h1>Miami</h1>
+    <div class="container container__header">
+      <header class="header">
+        <nav class="navigation">
+          <a class="navigation__items" href="#">
+            <img class="navigation__logo" src="./images/logo.png"
+              alt="logo nothing(r)"></img>
+          </a>
+          <div class="navigation__icons">
+            <a class="phone" href="tel:+0">
```
mvjl000, line 22: Add the same phone number as in the span so it's consistent ;)

mvjl000 (changes requested):
1. consitent href for tel: links
2.  As I user, I would expect close menu button to be in the exact same position as the open menu button
Screenshot 2023-07-19 at 13 18 08
Screenshot 2023-07-19 at 13 18 14

3. Some weird margins appear on smaller devices
Screenshot 2023-07-19 at 13 21 26
Screenshot 2023-07-19 at 13 22 54

src/index.html
HaniaNassalska: yeah, I did it on purpose (like with address before) - the phone number is fake, so what is a difference?

choeqq (changes requested):
Please fix margins on mobile and you're good to go 
image

PR #2543, opened 2023-07-12
Develop

Sancz0pansa:
[DEMO LINK](https://Sancz0pansa.github.io/layout_miami/)


choeqq (changes requested):
1. Content margin on mobile should be smaller 
image
2. On the other hand I think content should take more space on desktop 
3. Email should redirect us to email window, so we can send email, currently we're navigated to home page
image
3. I think menu popping animation is little too slow, we don't want to make our users wait 😄 

Other than that, looks really good! 🚀

choeqq (approved):
Looks good to me, nicely done 🚀

PR #2546, opened 2023-07-12
Add task solution

klaudia-klis:
[DEMO LINK](https://klaudia-klis.github.io/layout_miami/)


JoyCoffeeAddict (changes requested):
1. Try to come up with some idea to make this text more readable e.g. search how to add text-stroke, maybe add some background
image

2. In the menu "call to order" should be some kind of action. Maybe also a phone call, or a link to form, experiment with this.

3. This link should take us somewhere
image

4. Add validation to form so that we can't post empty request 

5. There are some resolutions where address and form are too close to each other.
image

6. Can't use phone from navbar: 
Peek 2023-07-18 11-47

7. This text is too wide on small screens 
image

8. Lower margins on smaller screens
image

Good job!

choeqq (changes requested):
1. Please provide form validation - if I don't type my email form should tell me why I can't send request
2. Address should take us to the place on Google Maps
3. Those should also be clickable
image
4. Main logo should always be a link to home page

choeqq (approved):
Nicely done ✅

PR #2548, opened 2023-07-13
Landing pages CATALOG

patryk-babon:
[DEMO LINK](https://patryk-babon.github.io/layout_miami/)


choeqq (changes requested):
That's crazy you've completed almost every single layout 😄 you nailed it! 🚀 

One little issue: menu shouldn't be a link - we don't want user to navigate to it, if we press previous in browser we don't want to be redirected to menu 👍 

Great job 🚀

choeqq (approved):
Good work, approved ✅

PR #2551, opened 2023-07-13
Develop

AndrzejPec:


choeqq (changes requested):
1. I can't close the menu, either on button nor logo, please make logo redirect to home page, and close button work
2. Get rid of side bars - use default background - as on section, and stretch hero image on desktop
image
3. Add form validation - we don't want to send empty requests
4. This section shold be interactive - [checklist](https://github.com/mate-academy/layout_miami/blob/master/checklist.md) 
image
5. `Book a test ride` should redirect us to phone call
image
6. Every menu link should have correct `href` attribute - [checklist](https://github.com/mate-academy/layout_miami/blob/master/checklist.md
7. Please fix horizontal margin on mobile - we don't want our page to shrink even more than it necessary, we want to have as much space on mobile as possible

AndrzejPec:
Introduced changes 3-6.
Close button and logo in menu both close the menu on my local copy, not sure why not working after deploy.
And unfortunately cannot go on today's Q&A, will have to work it out tommorow. So no change to 1, 2 and 7 yet.

mvjl000 (changes requested):
1. The header with it's background-image should take entire available space, on all devices.
Screenshot 2023-07-20 at 13 28 21

2. Menu button doesn't close the menu. Furthermore, as I user I would expect the open button and the close button to be in exact same position

3. Display the phone number on phone icon hover
4. Your links have `cursor: unset;`. I see that it ts a matter of css reset but I would change it to pointer ;)
5. Please do something about inputs/buttons border. Some default styles are applied which causes in some grayish color.
Btw - you should use Poppins font in inputs as well ;)
Screenshot 2023-07-20 at 13 34 41
6. I believe that links in the menu should have underline effect

choeqq (changes requested):
Waiting for any updates @AndrzejPec 👀

choeqq (changes requested):
Hero image should take the whole viewport, please fix that and you're good to go 🚀 
image

src/scripts/main.js
```diff
@@ -1 +1,76 @@
...
+    link.addEventListener('click', function(event) {
+      event.preventDefault();
+
+      const target = document.querySelector(this.getAttribute('href'));
+
+      target.scrollIntoView({
+        behavior: 'smooth',
+      });
+    });
+  });
```
JoyCoffeeAddict, line 32: You can achieve this without JS. Because now the page does not work as intended.

JoyCoffeeAddict (changes requested):
1. Clicking on the menu link should close the menu
Peek 2023-07-24 10-23

2. Add visible validation to form.

3. Form and address should be closer on tablets
image

JoyCoffeeAddict (approved):

PR #2555, opened 2023-07-14
add menu solution

MaciejKujawa:


choeqq (changes requested):
Please provide DEMO LINK, so we can view your work 😄 

Also you need to add that in readme.md 🚀

choeqq (changes requested):
1. Due to textfield overflowing, browser adds horizontal scroll for the whole page, please fix that
image
2. This section shouldn't be growing as big on desktop
image
3. Please keep your styles consistent - textfield is different than 2 other fields
image
4. Menu should have little bit more left margin 
image

choeqq (changes requested):
Waiting for any updates @MaciejKujawa 👀

choeqq (approved):
Looks good, approved ✅

PR #2562, opened 2023-07-15
Develop

emil-owczarek:
[DEMO LINK](https://emil-owczarek.github.io/layout_miami/)


choeqq (changes requested):
1. every phone number / address / email - should redirect us to correct service - [checklist](https://github.com/mate-academy/layout_miami/blob/master/checklist.md) 
2. This section should be visible when hovering over phone number
image
3. Images in this section are growing too big on desktop
image
4. This button shouldn't be that stretched
image

4. Please add form validation - we don't want to send empty requests, also according to checklist - we don't want to reload page on form submit
5. This section should also be interactive
image
6. I think mobile menu should take whole viewport height
image
7. Button text should be centered
image

choeqq (changes requested):
Waiting for any updates @emil-owczarek  👀

choeqq (approved):
Well done 🚀

PR #2570, opened 2023-07-18
add task solution

arturwisnia:
[DEMO LINK](https://arturwisnia.github.io/layout_miami/)


JoyCoffeeAddict (changes requested):
1. Add more spacing between columns, and change the proportion of fields to text on the right side.
image
2. Change background color from `#000`, there is too much contrast with text 
3. Prevent default behavior on sending message, I'm redirected to 405 page
4 Add padding, also close menu button should be on the same place as opening menu button
image
5. Menu links should close the menu 
6. It shouldn't be possible to scroll while using menu:
Peek 2023-07-24 10-23
7. Menu and phone icons are too big, but if you insist then you can keep it this way.

mvjl000 (changes requested):
Please apply other mentors' suggestions and try to make the website's design more similar to what is in figma

mvjl000 (changes requested):
1. Size of things: different size of open/close menu icons, enormous input fields, explore button should take full section width only on mobile,
2. 'Move Free' heading wrapping,
3. Font weights are a bit off,
4. In contact form send button should have the same width as input fields,
5. Why is the menu white, while in the figma it's greyish,
6. Menu links should be uppercase,
7. Bigger margin between contact form and footer,
8. Open/close menu buttons must have the same position,
9. Paddings/margins on the entire website (menu as well)

choeqq (changes requested):
1. Sizing: 
image
image
image
Please make sure you're trying to imitate Figma project
2. Paddings and margins around the content are little off

choeqq (approved):

PR #2576, opened 2023-08-02
Develop

LukashevychSergey:
- [DEMO LINK](https://LukashevychSergey.github.io/layout_miami/)


ivbutsykin (changes requested):
Hey, I can reproduce some problems from checklist: https://github.com/mate-academy/layout_miami/blob/master/checklist.md

Please, fix them:

4. Change text color on hover for phone, email and address
5. When you click on phone icon or phone number in contacts section, make sure that there is no 404 error, make it a real link to start a call on device
7. When clicking on any location / address - prevent errors and make it to open location in Google Maps
9. Location-related addresses / links should open google maps in a new tab target="_blank"
13. Placeholders in the forms suggest what to enter; apply validation of the form fields (required, email / tel etc.), then it is clear in what format to enter the data
14. Form shouldn't be submitted if some of the fields are not filled
15. Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)

DarkMistyRoom (changes requested):
- add a number on the hover
image

- these should be tel links
image

- fix this overlapping jump effect(https://www.loom.com/share/b4aeb59477794c729fe6d3189b58927f)
- make less scale to avoid text overlapping
- check the button text
image

- side indents should be equal throughout the whole page
image
image
image
image

- it should be a link
image

- make the cursor a pointer over the buttons

- clear the form on submit
- check the border-radius for the textarea
- move the background a bit to avoid icons and hair overlapping as they have similar color
image

- add a hover effect to these links
image

DarkMistyRoom (changes requested):
- the number should be reachable. But when I move cursor, it disapears
image

- add hove effects to these links too
image

- the right should be the same as the left one. Use `container` to limit the width of content according to the mockup. Max width should be 1260px
image
image

- the hover effect should be smooth in both directions for these elements as well as for all other
image

- it looks like the header's background problem is because you do not use full-size image for it
- center the form on the mobile
image

- add scroll to the menu to make it works fine in landscape mode
image

DarkMistyRoom (changes requested):
You're almost there!

- Make the hover effect smooth for this link in the menu
image

- make the hover effect smooth in both directions
image

- keep 20px side indents on a mobile throughout the whole page
image
image
image

- The top indent looks too big in the landscape mode
image

DarkMistyRoom (changes requested):
Much better! Fix these two moments:
- check this indent. It looks bigger in the mockup (desktop, tab)
image

- the hover effect stops working for the button after clicking on it
image

DarkMistyRoom (approved):
Well done!

PR #2591, opened 2023-10-11
Develop

Krzychu92:
krzychu92.github.io/layout_miami/


volodymyr-soltys97 (changes requested):
This link doesn't working( 
You need to copy the entire link and paste it here
image

Krzychu92:
> This link doesn't working( 
> You need to copy the entire link and paste it here
> image
> 

Updated

Krzychu92:
image
I can't sort out why the layout is not scaling down :/

Krzychu92:
@loralevitska that is a mobile version thats why I havent done any hover actions. The task says to send PR after each block is done so I sent a mobile version to check if its okay to keep going xd

etojeDenys:
> @loralevitska that is a mobile version thats why I havent done any hover actions. The task says to send PR after each block is done so I sent a mobile version to check if its okay to keep going xd

next time, write briefly about what you implemented

etojeDenys (changes requested):
good job, keep going

1. disable scroll when menu is open
image


2. the button should look like this on hover
image

Krzychu92:
Added a tablet and desktop version. Please check

volodymyr-soltys97 (changes requested):
1. Add a favicon and actual title on the page
image

2. Need to remove the scroll when this menu is open
image

3. The prices should be here on the same line
Demo
image

Design
image

4. This image should be the full width of the browser window
Demo
image

Design
image

5. The user will not be able to change the size of this field
image

6. Add cursor pointer and transition for hover effects here
image

7. Make all these fields are required before sending
image

8. These elements should be clickable
image

9. Need to remove autocomplete styles and fix the indents for the textarea
image

Krzychu92:
I think I have changed everything from list. I couldn't find how to block scrolling without JS so I setup menu over a layout with z-index

nazarmatsevych (changes requested):
Looks great, let's make some improvements:

1. Page should not reload on submit, also clear form fields on button click (
add this to your form 
image
)
2. Open external link like googlemap in a new tab
image

Krzychu92:
done it. Now I think is good :D

nazarmatsevych (approved):
👍

nazarmatsevych (approved):
Awesome 👍

PR #2594, opened 2023-10-20
BOSE landing page

zaura333:
[DEMO LINK](https://zaura333.github.io/layout_miami/)


zaura333:
For now it's only styled for mobile (320px)

zaura333:
Now I have to work on menu, styles for buttons and inputs, then I'll make it responsive

JoyCoffeeAddict (changes requested):
You page should work one every mobile device, so you can't just go: "it works on 320px mobile" 
Using hard-coded values for things like navbar is a bad idea: 
`.header {
grid-template-columns: repeat(2,130px);}
`
it will make your life harder later

mvjl000 (changes requested):
1. Use the exact same color for phone email and address that is used in the figma 
Screenshot 2023-10-23 at 08 53 19
2. Try to disable text area resizing and make sure that all corners are rounded
Screenshot 2023-10-23 at 08 54 26

JoyCoffeeAddict (changes requested):
1. Add a favicon
1. Center the text vertically
image
1. I would make the phone number clickable here"
image
1. Checklist points 4, 8, 15, 18 

Other than that, I think the page looks good. Tag us here or ping us on the slack if you don't understand any of the issues.

zaura333:
I made favicon out of sound waves image as the logo isn't really fit for that use, can change back if needed

Also I changed articles in recommend and categories sections into links. Texts scale alongside the images because I thought it might make sense in the context of this page. If that's not right for the task, I'll restructure that as well! 😀

JoyCoffeeAddict (changes requested):
1. There is some horizontal scrolling on my screen
image

1. Text scaling is not a bad idea,but I think that only one image should scale on hover
image


Some small adjustments and you're done! I think that the page is great.

mvjl000 (approved):
This looks great! 🚀

PR #2598, opened 2023-10-23
mobile layout without menu.

adrian200002:


JoyCoffeeAddict (changes requested):
1. Images here should be centered on every mobile device:
image
1. The same here 
image
1. Button should be stretched for the whole page's width 
1. The same about his image 
image
1. The same thing about the form 
image

Remember: Responsive means that our page looks good on every resolution, not only on resolutions picked by the designer. Your site should look good on every mobile, not only the 320px width ones.

JoyCoffeeAddict (changes requested):
1. Prevent sending the form as described in the checklist. Now it sends the form and takes me back to the top of the page. 
1. Address should be working: 
phone number should start a phone call 
email should open an email app
Try to google how to do that first. 
image

Your mobile version is pretty nice, make navigation for mobile and keep working on tablets and desktop version.

mbruhler (changes requested):
Good job so far! Your mobile version looks great.

Start working on other devices views

mvjl000 (changes requested):
- Open/Close menu button should be in the exact same position,
- There should be a phone number visible on phone icon hover,
- Nav icons should be a bit bigger (for tablet and desktop),
- The menu should be gray, not white,
- Phone number in the menu should also be a callable link,
- You should have some free space at the bottom of the page,
- Form submit shouldn't reload the page,
- The Textarea should have the same font family as other inputs

mbruhler (changes requested):
Hi,
Good job so far!
Couple of changes are needed:

Your form has incorrect font when filled

https://github.com/mate-academy/layout_miami/assets/21124163/f25a684f-9302-4632-906f-0abf65c710fe

Your send button has incorrect styling (there is shiny texture on top). Propably you have to reset the default button css
image

Your top section should strech the content to viewport size. Now the phone number on hover goes on the dark part of the image and everything mixes together and it's hardly visible. Move the image or stretch the content to max size of the screen.
image

When done it will be finished :)

mbruhler (approved):
Approved

PR #2601, opened 2023-10-23
develop

michal-reszka:
https://michal-reszka.github.io/layout_miami/


JoyCoffeeAddict (changes requested):
1. Paddings in the navigation are not the same as in the figma
image
1. Phone icon on hover does not work 
1. Smooth scrolling is not working 
1. Wrong font-family here 
image
1. Hovering images does not work 
1. Some styling mistake with a button 
image
1. Form validation as described in the checklist

mbruhler (changes requested):
Hi,

On mobile menu close button is in bad position
image

On mobile, Explore button is very small
image

On mobile, grid items are badly positioned 
image

On desktop, logo is blurry and logo items are badly positioned
image

Your contact form has extra outline which is not compliant with figma mockup
image

src/styles/blocks/details.scss
```diff
@@ -0,0 +1,138 @@
...
+  .details_paragraf {
+    width: 672px;
+    height: 66px;
+
+  }
+  .details_button {
+    width: 276px;
+    &:hover {
+      cursor: pointer;
+      transition: box-shadow 0.3s;
```
mvjl000, line 122: The transition mustn't be inside a hover. That's why your transitions are not smoothly going back to initial styles when you stop hovering them. Is should be moved up.

mvjl000 (changes requested):
- form: send button has default borders (remove it), form submit shouldn't reload the page, simply clear the form values, I would disable resize for the textarea,
- Open menu icon should have the same size as the phone icon (e.g. on the desktop it is way to small),
- Open/Close menu button should be in the exact same position,
- On the tablet resolution, the "phone" text shouldn't be higher that the input
Screenshot 2023-10-27 at 14 49 00

src/styles/blocks/details.scss
mvjl000: To be clear - you should do so for all transitions in order to have them nicely 'animated'

mvjl000 (changes requested):
I want you to do something about your hover effects. As I said noticed in the previous review, you are applying transition inside of `hover` block. It has to be outside of it in order to have smooth 'exit' animation.
Speaking of hover effects, I would like you to prevent the images (for example from 'compare bike') from overlapping on text underneath them.

PR #2602, opened 2023-10-23
Develop

jacekkosiba:
[DEMO LINK](https://jacekkosiba.github.io/layout_miami/)


mvjl000 (changes requested):
1. There should be a phone number visible on phone icon hover,
2. I think that the phone number in the menu should also be a callable link,
3. Speaking of the menu - maybe you could add some animation? (like slide in or something),
4. The cursor should be pointer on open/close menu button hover. It's good to 'inform' the user that some element is clickable,
5. Those should have the same height
Screenshot 2023-10-24 at 12 33 44
6. The buttons are missing a hover effect, same thing with inputs. Check inputs & buttons in figma

mvjl000 (changes requested):
1. As I have pointed out in the previous review - images from the recommended section should have the same height,
2. Checklist points - 1, 4, 5 (for mails - mailto:), 7, 8, 9, 10, ...- take a look at other points as well (especially form related ones),
3. The send button should have the same styles as the other button at the page, in the figma it's just shown in the hover state

jacekkosiba:
Sorry about last PR - I did not notice the problem with the improper images heights because it seemed to occur only on Firefox. I will remember to check my styles on more browsers from now on, I promise! :)

mvjl000 (changes requested):
- There is a strange effect on some of your images hover (All products, audio, accessories).  While hovering on just one image, both of them zoom in,
- Your navlinks are missing underline on hover

I appreciate some details you have made 👌

jacekkosiba:
Hi! Thanks for the feedback! But I feel that have to defend my code  :D
1.  The strange images hover effect that you mentioned is actually what I wanted to achieve - what zooms in is a category off products, one category contains two images, therefore two images zoom in on hover.
2. The navlinks have underline on hover, but it's working above 1260px of screen width (I just assumed that the underline is relevant only for bigger screens). I can see how this may be misleading, I will fix it in a minute

mvjl000 (approved):

PR #2608, opened 2023-10-25
mobile done

aureliagorzna:
- [DEMO LINK](https://aureliagorzna.github.io/layout_miami/)


mvjl000 (changes requested):
I suppose that your solution is only for mobile devices for now. However it shouldn't have such a fixed values that it looks off on just a bit bigger mobile device. Your solution is tailored for 320px width and it's not a good way to handle RWD. Please try to improve it.

Besides of that:

- No favicon
- Wrong website title
- Menu doesn't open

mvjl000:
Also please remember to get through the checklist and make sure that you implement all the requirements

mbruhler (changes requested):
Hi,
For now changes are:
- page should fill the viewport on > 1260px viewports
- form should not reload page and should reset
- phone should show phone number
- menu doesn't work
-

mvjl000 (changes requested):
I can't see any changes since previous reviews, please take not of our comments.

JoyCoffeeAddict (changes requested):
Resolve previous issues.

mbruhler (changes requested):
Your landing page is not compliant with the mockup and checklist. Please read it again carefully and style your website properly according to mockup.

mbruhler (changes requested):
There are still issues
- 744px viewport is wrong 
image
- 1260px viewport is wrong
image
- letter spacing is wrong on phone number
- header should be 100vh and properly sized so it's compliant with figma (you can't achieve proper alignemnt without 100vh) NOTE: despite mobile
- left padding on mobile is less than right 
image

JoyCoffeeAddict (changes requested):
1. Email link is not working (should open email app)
image
Validation in the form is not working correctly 
Animation
Look at the example from [documentation here](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation)
image

It is way better than before, so good job!

JoyCoffeeAddict (approved):

PR #2609, opened 2023-10-26
MyBike

Timi2424:
[DEMO LINK](https://Timi2424.github.io/layout_miami/)


mbruhler (changes requested):
Hi,
Good job on mobile version. There's still lots to do for other viewports

Also on mobile, last image (the bottom one) is too big.

mvjl000 (changes requested):
- Textarea should have the same font as other inputs (currently it's monospace), send button shouldn't have default border, values of the form should clear on send
- You are missing the smooth scrolling,
- Both of these should be a call-able link 
Screenshot 2023-10-27 at 15 00 59
- Speaking of links (e.g. in the navigation) - they should have cursor pointer. All links should have cursor pointer.  In my opinion also clickable elements like buttons should have it too. It tells the user that he can interact with it. Maybe you would consider it? Tell me what you think ;)

Timi2424:
> * Textarea should have the same font as other inputs (currently it's monospace), send button shouldn't have default border, values of the form should clear on send
> * You are missing the smooth scrolling,
> * Both of these should be a call-able link
>   Screenshot 2023-10-27 at 15 00 59
> * Speaking of links (e.g. in the navigation) - they should have cursor pointer. All links should have cursor pointer.  In my opinion also clickable elements like buttons should have it too. It tells the user that he can interact with it. Maybe you would consider it? Tell me what you think ;)

Timi2424:
from my side, the textarea has proper font family and send button has box shadow on hover

mvjl000 (changes requested):
- Smooth scrolling, contact link (from menu) should head the user to the contact section

mvjl000 (changes requested):
On tablet resolution the header title should be way lower
Screenshot 2023-10-31 at 08 43 46

You are still missing cursor pointer for the links (see one of my previous reviews)

src/styles/blocks/body.scss
```diff
@@ -0,0 +1,38 @@
...
+      opacity: 1;
+      transform: translateX(0);
+      pointer-events: auto;
+    }
+  }
+}
+
+a {
+  &:hover {
+    cursor: url(../images/cursor/cursorhover.png), auto;
```
mvjl000, line 36: Simply do:
```suggestion
    cursor: pointer;
```

mvjl000 (changes requested):
About those cursors...

src/styles/blocks/body.scss
mvjl000: By the way, it shouldn't be inside the hover {}

mvjl000 (changes requested):
`cursor: pointer;`

mvjl000 (approved):

PR #2615, opened 2023-11-17
MET landing page

usernameluke:
https://usernameluke.github.io/layout_miami/


lerastarynets (changes requested):
Pls watch the video review [via the link](https://www.awesomescreenshot.com/video/22605154?key=6e0578127fd1131c412a2a629c79e541) :)

witflash (changes requested):
Hey!
Great work!

I found several problems which you can fix.

1. You didn't add links for both tel numbers, but only to the icon. But how the user can directly call the second one? You have to add links to both numbers.
image
image


2. It would be great to add a fav icon. I see you haven't it yet.
image

3. Add HTML5 form validation to show errors to the user:
- required field
- email validation
image

4. In addition, reset the form on submit button click without page reload.

PR #2617, opened 2023-12-06
Develop 1

Laisens:
[DEMO LINK](https://Laisens.github.io/layout_miami/)


volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your page
1. Need to fix the background image position as on the design
Demo
image

Design
image

2. Need to fix the phone number color as on the design
image

3. Need to remove the scroll when this menu is open
image

4. Add a scroll behavior smooth on the page

5. Add hover effects for images
image

6. Need to remove the autocomplete styles for the inputs
image

7. Make all these fields are required before sending, also, when you click on send, you don't need to refresh the page, it's enough just to clear the fields
image

8. Make resize none for this field
image

9. These elements should be links
image

volodymyr-soltys97 (changes requested):
To improve:
1. When hovering, one image should not overlap another
image

2. Add transition for this hover effects
image
image

3. Add cursor pointer for this button
image

4. Still not fixed from the previous review, need to make all these fields are required
image

5. Add hover effects for all clickable elements
image

6. This element should be link on the burger menu
image

nazarmatsevych (changes requested):
1. Add required attribute to your form fields
image

2. Add transition to buttons and all elements with hover effects

If you have any questions - feel free to ask them in fe_chat

Laisens:
> 1. Add required attribute to your form fields
> 
> 2. Add transition to buttons and all elements with hover effects
> 

All the specified attributes are on the page, I don’t understand what else I need to add

anastasiiavorobiova (changes requested):
Great work! Almost done!
Screenshot 2023-12-10 at 15 40 57
Consider adding a type submit here 
Screenshot 2023-12-10 at 15 38 31
Consider adding a proper href attribute here to make a user able to call 
Screenshot 2023-12-10 at 15 43 59
Consider fixing the validation error

volodymyr-soltys97 (changes requested):
Almost done!
To improve
1. Need to hide the phone number on the mobile version
image

2. Why did you remove the autocomplete for the form?
It does not need to be excluded but only you need to stylize
Here you can read how changed the styles of the car
https://css-tricks.com/snippets/css/change-autocomplete-styles-webkit-browsers/
image

3. After the click on the button, the Page Sent should not be updated and should not be transferred to the beginning of the page, the fields should only be cleaned and all, also the text entered does not need to be displayed in the URL address
image
image

volodymyr-soltys97 (changes requested):
Almost done!
Need to make all these fields are required before sending
image

volodymyr-soltys97 (approved):
Well done 🔥

PR #2626, opened 2024-01-27
Develop

aMtiIV:
[DEMO LINK](https://aMtiIV.github.io/layout_miami/)


anastasiiavorobiova (changes requested):
Great work! Keep doing!

volodymyr-soltys97 (changes requested):
Good job 👍 
To improve:
Need to implement the burger menu when user click on this icon
image

Keep going with the next sections

src/index.html
```diff
@@ -1,13 +1,321 @@
...
+          <div class="message__item">
+            <label class="message__label" for="name">Name</label>
+
+            <input
+              class="message__input"
+              id="name"
+              name="name"
+              type="text"
+              placeholder="Name"
+              pattern="[A-Za-z]"
```
yevhenii-pyl, line 264: Good job! But we can improve form validation here: 

This pattern only allows single-letter input. If you want to allow multiple letters, you can modify the pattern or use + to allow one or more occurrences:
`[A-Za-z]+`

Also, we can improve email validation with something like this:
` pattern="[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}" `

yevhenii-pyl, line 264: Good job! But we can improve form validation here:

This pattern only allows single-letter input. If you want to allow multiple letters, you can modify the pattern or use + to allow one or more occurrences:
[A-Za-z]+

Also, we can improve email validation with something like this:
pattern="[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}"

yevhenii-pyl (changes requested):

yevhenii-pyl (changes requested):
Good job! You have done it correctly and in accordance with the provided design.
But we can try to improve it a bit: 

As you can see, on a wider screen resolution I'm losing your header buttons ( they are hidden in the hair :) ). 
What if we change columns in the header from being `repeat(...63px)` to => `repeat(..., 1fr)` ? 

Also, this will require you to add padding on the sides for 120px for the header 
vide
.

yevhenii-pyl (approved):
Good job! Looking good to me.

PR #2628, opened 2024-02-08
Develop

JustinNowoczek:
I had some issues with the commit and push (I accidentally did “commit and amend” on the lint fix instead of a regular commit) should be okay now (?)
I have done the header of the MyBike landing page, at the smallest size (mobile first), and the readme says to get it checked out every block we make, so here I am. Fingers crossed that I didn't mess up.

  - [MYBIKE Figma](https://www.figma.com/file/NZQAIydtHo5QkINyGLHNcq/BIKE-New-Version?node-id=0%3A1)

  - [Deployed Demo](https://tidetwo.github.io/layout_miami/)
  
  You will need to use devtools to see the mobile version properly, the page was designed for a phone with:
  ```width: 320px height: 568px```
  to match the figma design.


nataliaklonowska (changes requested):
Great work 🔥 
I suggest you slightly change the naming of your blocks. Header should be the part with logo and navigation, not the whole front page :)

TideTwo:
I did not mean to request review yet, but I'm not sure how to cancel, so I'm leaving a not here. I still have some issues to fix.

src/index.html
```diff
@@ -3,11 +3,223 @@
...
     <script src="scripts/main.js" defer></script>
+    <link
+      rel="icon"
+      type="image/x-icon"
+      href="./images/favicon.svg">
   </head>
-  <body>
-    <h1>Miami</h1>
+  <body class="page">
+    <menu class="page__menu menu">
```
asakevych, line 15: no need to use menu here, it's navigation, so nav would be enough

src/index.html
```diff
@@ -3,11 +3,223 @@
...
+      <a href="#" class="menu__link">About Us</a>
+      <a href="#" class="menu__link">Compare Bikes</a>
+      <a href="#" class="menu__link">Details</a>
+      <a href="#" class="menu__link">Contacts</a>
+      <a href="tel:+1 234 5555-55-55" class="menu__link menu__link--phone">
+        +1 234 5555-55-55
+      </a>
+      <a href="#" class="menu__link menu__link--book">
+        Book a test ride
+      </a>
```
asakevych, line 42: create list for those elements

src/index.html
```diff
@@ -3,11 +3,223 @@
...
+            Send
+          </button>
+        </form>
+        <div class="contact-us__info contact-info">
+          <h2 class="contact-info__header">Phone</h2>
+          <a href="tel:+1 234 5555-55-55" class="contact-info__link">
+            <p class="contact-info__desc">+1 234 5555-55-55</p>
+          </a>
+          <h2 class="contact-info__header">Email</h2>
+          <p class="contact-info__desc">hello@miami.com</p>
```
asakevych, line 209: it should be <a> with possibility to send email

asakevych (changes requested):
So far so good!! :rocket: 
for me menu doesn't work
remember about semantic

choeqq (changes requested):
There should be form validation - I should get error message when trying to send empty form 
Screenshot 2024-02-13 at 15 26 02
In this section bikes should be aligned horizontally on wider screens

TideTwo:
@choeqq Could I please get an explanation on what to do? The validation on the form is disabled, (even though I use the attribute "required" and specific input types like "email" because I use preventDefault on the form, as suggested in the checklist. 
``` 15. Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) ```
Do I implement my own validation in js, or ignore that step? Also, on the second point, I checked with an extension, and the element seems to be centered for me? The page is already at the widest it can be on my monitor.
image

choeqq (changes requested):
@TideTwo form validation is fine right now - I get an error when trying to submit empty form 👍 

and when it comes to compare section:
you're right that the elements are aligned center, but when you'll take a look at figma design (desktop version) the bikes are side by size - horizontally, and despite the fact we're on desktop your solution still looks like mobile

TideTwo:
> @TideTwo form validation is fine right now - I get an error when trying to submit empty form 👍
> 
> and when it comes to compare section: you're right that the elements are aligned center, but when you'll take a look at figma design (desktop version) the bikes are side by size - horizontally, and despite the fact we're on desktop your solution still looks like mobile

The validation works now as i removed preventDefault and redeployed the project. I will get to implementing the proper layout now.

PR #2630, opened 2024-02-08
 Created landing page

YarynaPuhach:


YarynaPuhach:
[DEMO LINK](https://YarynaPuhach.github.io/layout_miami/)

choeqq (changes requested):
Screenshot 2024-02-09 at 08 51 08
Each element of this section should trigger an action - redirect to location on the map, start a call on your computer, open email app to send an email, just like the one in header

src/index.html
```diff
@@ -2,12 +2,356 @@
...
+    <header
+      class="page__header header"
+      id="header">
+      <div class="header__nav ">
+        <img
+          src="images/Logo.svg"
+          alt="Logo"
+          class="header__nav__logo">
+
+        <div class="header__nav__buttons">
```
nataliaklonowska, line 42: element should belongs only to a block so I could be header__nav-buttons or header__buttons [Typical BEM Mistakes
](https://mate-academy.github.io/fe-program/css/typical-bem-mistakes-en#an-element-of-another-element)
remember to fix it in whole project not only here ;)

src/styles/blocks/phone.scss
```diff
@@ -0,0 +1,21 @@
+.phone {
+  display: flex;
+  gap: 20px;
+  align-items: center;
+
```
nataliaklonowska, line 5: to remove underline you need to add text-decoration: none to phone class because it comes from tag a

nataliaklonowska (changes requested):
Great work 🚀 
firstly, correct class naming

src/index.html
```diff
@@ -2,12 +2,359 @@
...
+
+    <div
+      class="menu"
+      id="menu"
+    >
+      <div class="menu__header">
+        <img
+          src="images/Logo.svg"
+          alt="Logo"
+          class="menu__header__logo">
```
asakevych, line 71: `menu__header-logo` or `menu__logo`

src/index.html
```diff
@@ -2,12 +2,359 @@
...
+      </h1>
+    </header>
+
+    <div
+      class="menu"
+      id="menu"
+    >
+      <div class="menu__header">
+        <img
+          src="images/Logo.svg"
```
asakevych, line 69: should be inside <header> tag

src/index.html
```diff
@@ -2,12 +2,359 @@
...
+      <nav class="menu__nav">
+        <ul class="menu__nav--list">
+          <li class="menu__nav--item">
+            <a
+              class="menu__nav--link"
+              href="#recommended">
+              Recommended
+            </a>
+          </li>
+          <li class="menu__nav--item">
```
asakevych, line 91: `menu__nav-item`

src/index.html
```diff
@@ -2,12 +2,359 @@
...
+          class="menu__header__link">
+          <img
+            src="images/Icon-Close.svg"
+            alt="Icon-Close"
+            class="menu__header__icon">
+        </a>
+      </div>
+
+      <nav class="menu__nav">
+        <ul class="menu__nav--list">
```
asakevych, line 83: it should be `menu__nav-link`
in BEM you have block, here it is your `menu`
Element is your `<ul>` so it can be `menu__list` or `menu__nav-list` 
`--` is for modifier, modifier is used to different variants of same element
for example if you have column list or row list you can add modifier `menu__nav-list--row` or `menu-nav-list---column`
but here is no need to use modifier

src/index.html
YarynaPuhach: Sorry doesn't really get it
Could You please explain

YarynaPuhach:
> Good job! fix BEM and make your menu not scrollable on the phone view

do You mean owerfolw-X: hidden;? cause in checklist 
"
  /* Add this rule to add scroll when there is not enough space */
  overflow-y: auto;

  ..."

choeqq (approved):
LGTM ✅

PR #2639, opened 2024-02-13
Develop

konradwasilewski444:


choeqq (changes requested):
paddings on logo and menu

compare bikes section should be horizontal on wider screen

images should be smaller in details

contact us form should be side be side with phone, email and address

page shouldn't reload when submitting the form

form should have validation - no empty forms allowed 

menu should work

phone, emial and address should be interactive (I should be able to call / send email after click)

konradwasilewski444:
As you can see i pulled request view hours ago, but then i made a new commit. Now i cant do pull request form vesrion 'almost final version, maybe some smallers things to do'.  This one is right one

choeqq (changes requested):
please checkout figma design, I think the desktop view looks little different, than the one you submitted
I shouldn't be able to send empty form, and when trying to do that, I should get an error message
I can't see bike pictures in compare section 🤔 
phone, email and address in contact section should be clickable, and trigger actions

konradwasilewski444:
"As you can see, I've corrected most of the errors, however, in the 'contacts us' section, the link redirecting to the email doesn't seem to be working. The address and phone number are more or less okay. I've searched the internet for ways to redirect it and found only <a href="mailto:..>. Is that a good way to do this ?

src/index.html
```diff
@@ -4,10 +4,252 @@
...
+            </h4>
+            <p class="address__content" id="number">+1 234 5555-55-55</p>
+            <h4 class="address__label">
+              <a href="mailto:hello@miami.com">Email</a>
+            </h4>
+            <p class="address__content">hello@miami.com</p>
+            <h4 class="address__label">
+              <a href="https://maps.app.goo.gl/uXQTCdJCemi7rnYj6">Adress</a>
+            </h4>
+            <p class="address__content">400 first ave.<br>suite 700<br>Minneapolis, MN 55401</p>
```
nataliaklonowska, line 245: phone number, mail or address should work as links not their labels ;)

nataliaklonowska (changes requested):
Your email redirection is working, but there are some other minor issues:
 
- checklist: Page shouldn't be reloaded on form submit (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)
- checklist: disable page scrolling under the menu
- phone number should be centered 

image

src/index.html
```diff
@@ -4,10 +4,263 @@
...
+      </nav>
+      <div class="links menu__links">
+        <ul class="links__list">
+          <li class="links__item"><a href="#home" class="links__link">Home</a></li>
+          <li class="links__item"><a href="#about-us" class="links__link">About Us</a></li>
+          <li class="links__item"><a href="#compare-bikes" class="links__link">Compare Bikes</a></li>
+          <li class="links__item"><a href="#details" class="links__link">Details</a></li>
+          <li class="links__item"><a href="#contact-us" class="links__link">Contacts</a></li>
+        </ul>
+        <div class="links__item links__item--number">+1 234 5555-55-55 </div>
```
asakevych, line 58: it should be link, not just div

asakevych (changes requested):
in figma image have smaller height and that's why it looks different here in mobile version
image
form is too small 
image
#about-us > section.main__section.main__section--3.section > div should have width:100%

src/index.html
```diff
@@ -4,10 +4,263 @@
...
+              id="message"
+              class="form__textarea"
+              placeholder="Message"
+              required
+              ></textarea>
+            </div>
+            <button
+            class="button button--primary"
+            type="submit"
+            id="button-primary">
```
asakevych, line 231: format code correectly
add names to your inputs

src/index.html
```diff
@@ -4,10 +4,263 @@
...
+              Send
+            </button>
+          </form>
+
+          <address class="address section__adress"> <!--Pomyślec o dobraniu w pary-->
+            <h4 class="address__label">
+              Phone
+            </h4>
+            <p class="address__content" id="number">
```
asakevych, line 240: why do you need additional p for links? can't you just style `<a>` element? 
it's a good practice to avoid html elements where they are not needed to prevent excessive-DOM size

asakevych (changes requested):
still need to positiob header image on mobile
image

konradwasilewski444:
I fixed problem with header background image in my last commit after I requested 'add solution' :)

choeqq (approved):

PR #2645, opened 2024-02-26
miami_layout task

Alex-qqren:
[DEMO LINK] (https://Alex-qqren.github.io/layout_miami/)


src/styles/blocks/header.scss
```diff
@@ -0,0 +1,12 @@
+.header {
+  display: flex;
+  flex-direction: column;
+  height: 100vh;
+  justify-content: flex-end;
+  box-sizing: border-box;
+  padding: 0 20px 40px;
+  background-image: url(../../../src/images/photos/1.jpg);
```
yevhenii-pyl, line 8: Check the path, there is no background image at your demo.
Also, you can use a shorthand for background property here.

yevhenii-pyl (changes requested):
That's a beginning 👍 

Check the path to images to display them correctly and add a link to the mockup in the description.

PR #2647, opened 2024-03-04
Develop

MelnykA:
[DEMO LINK] 
https://MelnykA.github.io/layout_miami/


sTorba24 (changes requested):
Good work! Lets improve it :)

1. cards shouldn't overlap each other
image


2. same here
image

 
3. check submit button styles 
image

4. form should be submittable, and reset after it

5. add cursor pointer to all buttons

6. here should be links

image

vadiimvooo (changes requested):
Good job, you are almost done. A few changes required:
1. Add validation for email field.
2. Use attribute targer="_blank" to open links in new tab.

etojeDenys (approved):
image
you don't need to open a new tab when the user clicks on this link

PR #2650, opened 2024-03-27
Mobile version ready

olafchuszno:
[DEMO LINK] (https://olafchuszno.github.io/layout_miami/)


yevhenii-pyl (changes requested):
Good job! I really like the way your commits are structured and commit messages - that may seem like an unnecessary part, but it's really important when it goes to a code style and clean code approach. 

Since your PR is called 'Mobile version ready' - I will check it only for the mobile. 

What can we improve:

1. We can refactor your approach with parent-child elements positioning. 
Here, you are using padding-top to position `navigation`, `links menu__links`.
`menu__paragraph` has a margin-bottom and `.menu__call-link` has a padding bottom. 
That's messy. As a general rule, a parent element should be responsible for positioning it's children. 
You can easily achieve it by adding a padding-top to  a parent element `menu` and using gap `property`.
Same goes for link and phone number - just wrap them with a div and use gap to position them. 
For the underline, I would suggest using text-decoration with text-underline-offset instead of border + padding.  
p

2. I would pretty much use the same approach with gaps instead of margins/paddings for other elements. 

Other than that, once again, good job! It looks really close to the expected result.
Keep pushing!

BogdanMaliuta (approved):
It`s really good!

PR #2652, opened 2024-04-05
Develop

n3di:
- [DEMO LINK](https://n3di.github.io/layout_miami/)


BogdanMaliuta (changes requested):
Hello here some things to fix:
Fix background image according to design 
image
Fix margins here
image
Bikes starts from uppercase latter
image
fix margin here
image
change font-famly here
image
fix this background image
image
Also check your mobile and tablet version according to figma design and fix

BogdanMaliuta (changes requested):
Hello, fix this: 
Footer Image 
image
Background in tablet version
image
Fix this on mobile version

yevhenii-pyl:
A bit concerned with this 'rainbow'. Why don't you just add padding-top to your header container? 

r

yevhenii-pyl (approved):
From the design perspective, it's pretty accurate. Good job! 👍 
We can and should move forward :) 

However, from the coding perspective there is some room for growth: 
1. You have columns provided at the design, why don't you use grids to position your elements? 
2. You tend to position elements 'just to make it work'. It works, but this is not the cleanest approach. As a common rule here, your parent (container) elements should be responsible for positioning it's children. It means, that, whenever possible, child element should never use margin | padding properties to position itself inside the container.

PR #2653, opened 2024-04-15
Develop

kgGugala:
- [MYBIKE landing](https://www.figma.com/file/NZQAIydtHo5QkINyGLHNcq/BIKE-New-Version?node-id=0%3A1)


yevhenii-pyl (changes requested):
Problem is fixed so let's move forward 💪 
You are doing a great job, can't wait for us to finish with this project and proceed to React learning path :)

yevhenii-pyl (changes requested):
gh

There is no `gh-pages` branch. Don't forget to deploy your work.

src/styles/blocks/navigation.scss
```diff
@@ -0,0 +1,32 @@
+@use '../utils/mixins.scss' as *;
+
+.navigation {
+  display: flex;
+  align-items: center;
+  padding: 24px 0 0;
```
yevhenii-pyl, line 6: I'd probably set this property as padding-top for the `header`. 

As a general rule, the parent container should be responsible for positioning it's children. Not a mistake here, but a possibility for cleaner approach.

src/styles/blocks/header.scss
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  &__title {
+    place-self: end center;
+    grid-column: 1 / 3;
+    grid-row: 2 / 3;
+
+    @include for-tablet {
+      grid-column: 1 / 7;
+      padding-bottom: 40px;
```
yevhenii-pyl, line 48: Maybe, it would make sense to position nav and title using grid-row property, but also not a mistake here, just a possible second approach.

src/styles/blocks/main.scss
```diff
@@ -0,0 +1,184 @@
...
+    grid-column: 1 / 3;
+
+    @include for-tablet {
+      grid-column: 1 / 4;
+      margin-top: 120px;
+    }
+
+    @include for-desktop {
+      grid-column: 1 / 5  ;
+      margin-top: 120px;
```
yevhenii-pyl, line 43: I'd create a separate container for this title && paragraph block to ease the positioning.

yevhenii-pyl (approved):
Really solid work! Appreciate the amount of effort you put here. 🥇 

Some things may be done just a bit cleaner, but they are minor. Other than that, your code is clean and readable. Good job!

PR #2654, opened 2024-04-16
Develop

jacube1:
[DEMO LINK](https://jacube1.github.io/layout_miami/)


BogdanMaliuta (changes requested):
Hello, some things to fix: 
fix header: 
image
fix recomeded: 
image
fix also this: 
image
image
image
Also check this all things on mobile and tablet versios

yevhenii-pyl (changes requested):
Solid! But we still have a few things to fix: 

Something is off with tablet mixin: 

tablet

desc

If you use padding-bottom to position elements - stick to it, mixing padding-top AND padding-bottom is considered to be a bad practice: 

p

About-us image should be fixed: 
img
(maybe, try to set it as a background property for a div, not an image)

yevhenii-pyl (approved):
LGTM 👍

PR #2655, opened 2024-04-23
Bose layout

WGoliasz:
Unfortunately, my project does not contain commits because at the end of the project I encountered a critical error that forced me to delete the entire directory and create a new one to which I moved the entire src folder.

[DEMO LINK] (https://WGoliasz.github.io/layout_miami/)


BogdanMaliuta (changes requested):
Hello, here some things to fix in your landing:
your burger menu, must be on this page,not another
image
Why its green?
image
Text must be in 1 line 
image
Text must be bold
image
Look at font 
image
Also check your mobile and tablet version.
Good luck!

BogdanMaliuta (changes requested):
Hello here some things to fix
tablet version: 
your left and right positions of logo, burger etc.
image
fix it according to design 
image
do same to all another page componets, also check your mobile version

WGoliasz:
Sorry, I didn't arrange the elements on the page carefully because I thought not to make it "pixel perfect".
Zrzut ekranu 2024-04-24 115317

WGoliasz:
I think it will be fine now :)

yevhenii-pyl (approved):
Solid work and definitely a good amount of effort put it here! 👍 

One thing I really want to notice here: you tend to use margin-top && padding-top to position your elements. And you do it for all elements consistently. Great job! 
It's a valid approach and if you started using `-top` property to position elements - you should stick to it and never switch to randomly using `-bottom`. 

Also, you could mark your sections as rows and position them using row-gap property. but that would be just another approach. Yours is also valid and do the job.

PR #2663, opened 2024-08-27
Develop

Bajkowsky:
[DEMO LINK](https://Bajkowsky.github.io/layout_miami/)


Zibi95 (changes requested):
Hey. Check the spacing and typography again for all the devices in figma.
Here are some things that i found: 
- The image in hero is no centered also resolution of image is lower than expected (quality is bad)
- This one text its weight is to high than in figma 
image
- the image overflows text when its hovered 
image

- contact form is not centered, its aligned to left in figma
image
- footer image is too big lower its height
- on mobile this image should also be centered 
image

- I know it don't have to be pixel perfect but some spacings between sections is off by tens of pixels

Zibi95 (changes requested):
Please follow the [checklist](https://github.com/mate-academy/layout_miami/blob/master/checklist.md) for the implementation. It seems you missed point number 15: the page should not reload on form submission. You can prevent this by using event.preventDefault() (https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault).

Upon inspecting the elements, I noticed you're using padding everywhere. Consider using the gap property in Flexbox to simplify your layout.

Additionally, the hero image is still not centered and appears to be of low resolution. Here is a [link](https://s3-alpha-sig.figma.com/img/75d2/97df/5101dbe1dd55f818092e4eb22a439eea?Expires=1725840000&Key-Pair-Id=APKAQ4GOSFWCVNEHN3O4&Signature=oWlKvMUzxxOtyxiXiAbuiC8TvzeMZZS5hxWCdm7kA~4jlcYC~ThzV7GAVTqwhhXrktkpFNibSpwlxYX5rUKYzRIn0RdDiR4nTTkdqznAteaGLgBGxCWZ0u5IW5l-3Dt02qeK0dep~UfoPWjfTS--TZwH9ZcWCfHZfejDKhhH5Jfc-OYIV8qf~UAp5NE1EvDjrag0xBCaaFbdgpEb53HldLt59687hl-WG~1fibCxiXVrFeshvZDsQBaWf6cAt2pcvKQiEtuihst55HqqLZnEUjFx1T6dfu2Glo4vzIsOcs~ASxdPsmOaDSDVEGJQtoOTYBf-FS~bkBfl8jPw3MM9EA__) to a higher resolution version.

Zibi95 (changes requested):
After recent changes layout for mobile is broken.
image

natalia-klonowska (changes requested):
- address link should open google maps in a **new tab** [[checklist 9](https://github.com/mate-academy/layout_miami/blob/master/checklist.md)]
- bike images in section compare bikes are poor quality 
- footer image should have different background-position (it's not centered) 
image

BogdanMaliuta (changes requested):
Hello, nice work, but need to make some changes, the most of them in tablet version 
change your font family for "message" 
image
change content in this header according to figma 
image
on tablet version change :
your background image position
image
styles here
image
image
image

Zibi95 (changes requested):
- Here spacing is to big you have applied correct values with padding and additionaly have gap 
image

- These images on tablet looks off 
image

- In many places, the issue with spacing is that you are applying both top and bottom padding, and these values are adding up.

Zibi95 (approved):
Hey, I can finally accept this PR, but I recommend refining it a bit more. It would be great if you could enhance it with some additional JavaScript features. Since it's a portfolio project, it needs to be both polished and representative.

PR #2673, opened 2024-12-26
MyBike

pat5513:
[DEMO LINK](https://pat5513.github.io/layout_miami/)


src/images/header-bg.jpg
natalia-klonowska, line 1: you should use full image as you can see on figma (its better quality and can adjust to different screen sizes): 
image
use background-position to hide sides of full image and get result like in picture you currently use ([here you can read more about how to use background-position](https://www.sitepoint.com/how-to-use-css-background-size-and-background-position/
))

natalia-klonowska (changes requested):
Great start 😃 remember to ask for another review after completing each section or block ;)

danon321 (approved):

PR #2676, opened 2025-01-05
Develop

o-koszalka:
https://o-koszalka.github.io/layout_miami/


danon321 (changes requested):
Tylko to widze na stronie
image

Nic wiecej :(

natalia-klonowska (changes requested):
header and title should adapt to the changing screen size. Currently they remain fixed in place across mobile, tablet and desktop while only the side margins change. Here’s how it should look on mobile just before switching to the tablet layout: 
image

o-koszalka:
> header and title should adapt to the changing screen size. Currently they remain fixed in place across mobile, tablet and desktop while only the side margins change. Here’s how it should look on mobile just before switching to the tablet layout: image

I understand what you mean, but could you maybe leave a hint on how i can do it using grid? Also I think that in Bartek's video related to this project, his site also "jumps" between mobile and desktop versions.

Zibi95 (changes requested):
Header looks fine. Code also for now looks ok and is well structured. Keep up the good work.

PR #2698, opened 2025-07-24
feat: add header section

Michal-Naron:


natalia-klonowska (changes requested):
zrób jeszcze raz deploy projektu i dodaj demo link do opisu pr

natalia-klonowska (changes requested):
layout is not responsive it's fixed. Even if you're currently working only on mobile version of design you should already be creating a layout that adjusts to different screen sizes.

PR #2706, opened 2025-08-27
task solution

Gabecznik:
https://gabecznik.github.io/layout_miami/


natalia-klonowska (changes requested):
- background image on tablet and mobile versions doesn't match design. title is positioned too low and icons should be spaced further apart (32px gap on tablet and desktop):
image

- Compare bikes section should have larger side margins and space between The details title and pictures should be 64px instead of 40px:
image

- typed text inside inputs shouldn’t touch the edges 
image

danon321 (approved):

PR #2707, opened 2025-09-03
Develop

ArturDrzazga:


natalia-klonowska (changes requested):
Great work! 🚀 one last thing to fix ;)

address section on mobile shouldn't be centered but on left and remove italic font style from address__label elements:
image

natalia-klonowska (changes requested):
Almost done! Just make sure to keep side margins consistent for this section:
image
Apply the same max-width as form to address element, set its width to 100% and center it using e.g. margin: auto to achieve this result:
image

PR #2708, opened 2025-09-12
Create header block

vlad-koval-8:
- [Demo Link] (https://vlad-koval-8.github.io/layout_miami/)


natalia-klonowska (changes requested):
your layout should be responsive but right now it's fixed because grid columns have fixed sizes. Grid shown in design is just an example and you don’t have to follow it exactly. You can make it responsive by using fr units or even build everything using only flex if you prefer. The choice of implementation is yours just make sure final result is fully responsive

danon321 (changes requested):
image

There underline on hover have fixed value. Try to do it on 100% of the string width. So every item in the list have matching underline width

Apart from that page looks great! Nice job :))

danon321 (approved):
No i super

PR #2725, opened 2025-12-24
Initial commit

Qbusik:
[DEMO LINK] (https://Qbusik.github.io/layout_miami/)


index.html
```diff
@@ -1,22 +1,401 @@
...
+        <div class="about__title" id="about-us">
+          <p class="about__accent">Plan your visit</p>
+          <h1 class="about__name">Museum hours</h1>
+        </div>
+
+        <article class="about__article">
+
+          <h2 class="about__title-2">The Met Fifth Avenue</h2>
+
+          <div class="about__item about__open-time">
```
natalia-klonowska: you only use `about__open-time` together with `about__item` so it would be better to make `about__open-time` a modifier of` about__item` e.g. `about__item--open-time`

natalia-klonowska (changes requested):
- ensure that no elements are on the same layer as menu as they remain visible during section navigation
image

- phone and burger icons should be bigger on tablet and desktop versions (24px instead of 18px)

image

index.html
```diff
@@ -1,22 +1,401 @@
...
+
+          <a href="tel:+1 212-923-3700" class="phone">
+            <div class="phone__label">
+              <p>The Met Fifth Avenue</p>
+              <p>+1 212-535-7710</p>
+              <p>The Met Cloisters</p>
+              <p>+1 212-923-3700</p>
+            </div>
+            <div class="header__phone"></div>
+          </a>
```
natalia-klonowska, line 43: [BEM:](https://mate-academy.github.io/fe-program/css/typical-bem-mistakes-en#using-an-element-inside-another-block) `header__phone` is nested inside `phone` block so it should use `phone__` prefix instead of `header__`

index.html
```diff
@@ -1,22 +1,401 @@
...
+            src="src/images/icons/logo.svg"
+            class="logo"
+            alt="logo">
+        </a>
+
+        <a href="#home">
+          <img
+            src="src/images/icons/close.svg"
+            alt="close"
+            class="menu__close">
```
natalia-klonowska: [BEM:](https://mate-academy.github.io/fe-program/css/typical-bem-mistakes-en#using-an-element-inside-another-block) `menu__close` is nested inside `navigation `block so it should use `navigation__ `prefix instead of `menu__`

natalia-klonowska (changes requested):

natalia-klonowska (approved):
Great work! 🚀

PR #2734, opened 2026-03-03
Bose landing page

damian-k-dev:
[DEMO LINK](https://damian-k-dev.github.io/layout_miami/)


natalia-klonowska (changes requested):
Great start :D 

- consider whether header really should be fixed because right now it blends in with the rest of page content which makes it hard to read. fixed header is usually used when it has some background and is clearly separated because content scrolling underneath can reduce readability what can negatively affect user experience. It’s important to ensure proper contrast and visual separation so that header remains clear and accessible

- remember to build responsive layout from the start. products in recommended section should be centered and/or scale proportionally as screen width changes

damian-k-dev:
> Great start :D
> 
> * consider whether header really should be fixed because right now it blends in with the rest of page content which makes it hard to read. fixed header is usually used when it has some background and is clearly separated because content scrolling underneath can reduce readability what can negatively affect user experience. It’s important to ensure proper contrast and visual separation so that header remains clear and accessible
> * remember to build responsive layout from the start. products in recommended section should be centered and/or scale proportionally as screen width changes

Thank you for the feedback :) I have updated header positioning and adjusted the recommended section so the products scale proportionally as the screen width changes

damian-k-dev:
> Thank you for the feedback :) I have updated header positioning and adjusted the recommended section so the products scale proportionally as the screen width changes

Ok my bad. After coming to the end of building my landing page I realised what you actually meant by saying "remember to build responsive layout from the start". At the beginning I thought it was only about centering the products, but now I understand that I should build responsiveness for each section from the start before moving to the next one.

Now I have to go through each section and adjust the responsiveness which is not ideal, but lesson learned. Next time I will definitely do it from the beginning.. Thanks for pointing this out :D

natalia-klonowska (changes requested):
- horizontal scroll appear on tablet version 

Image

- here you can see how hover and pressed states for buttons and inputs should look:

Image

- contact us section looks off between tablet and desktop breakpoints. button should scale proportionally with inputs to maintain visual consistency and to reduce too large right margin consider increasing gap between form and address

Image


[CHECKLIST:](https://github.com/mate-academy/layout_miami/blob/master/checklist.md)

- Page shouldn't be reloaded on form submit
- disable page scrolling under the menu

danon321 (approved):
Dobrze to wygląda. Dobra robota :))

damian-k-dev:
Dziękuję za sprawdzenie i ocenę! 🙂

PR #2741, opened 2026-03-20
MyBike page

Darmi555:
[DEMO LINK](https://Darmi555.github.io/layout_miami/)


natalia-klonowska (changes requested):
your demo link doesn't work. Please deploy your project using `npm run deploy` command

natalia-klonowska (changes requested):
- buttons should have the same font size and resize in the same way when screen width changes
Image

- sometimes a small gap appears between sections so it would be better to set background color for main instead of section

Image

- phone number should be displayed in single line if there is enough space

Image

[CHECKLIST:](https://github.com/mate-academy/layout_miami/blob/master/checklist.md)
- Change text color on hover for phone, email and address
- When you click on phone icon or **phone number in contacts section**, make sure that there is no 404 error, make it a real link to start a call on device
- When clicking on any location / address - prevent errors and make it to open location in Google Maps
- apply validation of the form fields (required, email / tel etc.), then it is clear in what format to enter the data

index.html
```diff
@@ -6,17 +6,174 @@
...
-    <title>Miami</title>
-    <link
-      rel="stylesheet"
-      href="src/styles/main.scss"
-    />
+    <title>MyBiKE</title>
+    <link rel="icon" type="image/png" href="photos/favicon.svg">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="anonymous">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@100..900&display=swap" rel="stylesheet">
```
natalia-klonowska: you should use Poppins font instead of Roboto

Image

natalia-klonowska (changes requested):
Great work! Just a few small fixes are needed: 

- images should be the same height

Image

- on desktop and tablet versions button should be shorter 

Image

index.html
natalia-klonowska: you’re still using Roboto font. change it to Poppins

natalia-klonowska (changes requested):

natalia-klonowska (changes requested):
Roboto is still applied in demo because css file hasn’t been updated. also remember that this project supports scss so there’s no need to include css files just update html to reference scss instead of css

Image

Darmi555:
The deploy page didn't work with SCSS in HTML, so I just updated the CSS and left it at that. I'm glad you said 'there is no need to include CSS files' rather than 'you can't include CSS files', right? :D

natalia-klonowska (approved):

PR #2749, opened 2026-06-25
Develop

MarekBiela:
Index without navigation
[DEMO LINK](https://MarekBiela.github.io/layout_miami/)


index.html
```diff
@@ -15,8 +15,229 @@
...
+      <h1 class="title">
+        Welcome
+        <br>
+        to the Met
+      </h1>
+    </header>
+    <main class="main">
+      <section class="hours">
+        <div class="hours--left">
+          <div class="main__font main__font--highlight main__font--highlight-small">
```
natalia-klonowska, line 33: [Typical BEM Mistakes:](https://mate-academy.github.io/fe-program/css/typical-bem-mistakes-en#using-an-element-inside-another-block) Using an Element Inside Another Block

index.html
```diff
@@ -15,8 +15,229 @@
...
+            Plan your visit
+          </div>
+          <h2 class="title title--secondary title--secondary-left">
+            Museum hours
+          </h2>
+        </div>
+        <div class="hours--right">
```
natalia-klonowska, line 40: [Typical BEM Mistakes:](https://mate-academy.github.io/fe-program/css/typical-bem-mistakes-en#using-a-modifier-without-the-belonging-class) Using a Modifier Without the Belonging Class

natalia-klonowska (changes requested):
Great start :) remember about [CHECKLIST](https://github.com/mate-academy/layout_miami/blob/master/checklist.md) 

- there're several BEM naming issues in your project. I've marked one example of each type of mistake but please fix all similar cases

- consider adding max-width so that on screens wider than 1260px only side margins continue to grow because right now some sections become too stretched (e.g in Contact Us section the two content columns are pushed to opposite sides of the screen leaving large empty space between them

- adjust image positioning so that it is cropped correctly

Image

- change cursor to pointer for all buttons and links (make sure all addresses are clickable links)

Image

Image

